PR #37, opened 2018-01-30
Change zone from hel1 to hel2

iler:


karvounis:
I see that the travis build is failing but this https://github.com/UpCloudLtd/upcloud-go-api/pull/38#issuecomment-372987797 can explain why.

PR #38, opened 2018-03-12
Restore previous timeout after server started

meafmira:


upcloud/client/client.go
```diff
@@ -3,12 +3,12 @@ package client
 import (
 	"bytes"
 	"fmt"
-	"github.com/blang/semver"
-	"github.com/hashicorp/go-cleanhttp"
```
Jalle19: Is this really not used anymore?

meafmira: @Jalle19 fixed

Jalle19:
@iler you should configure credentials in Travis CI so the tests can be run

iler:
@Jalle19 as this is a PR outside of the organization it does not allow the use of secret environment variables in the tests as explained here https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions. So those are configured there but can not be used in this PR.

Jalle19:
@iler ah, didn't know about that restriction!

PR #42, opened 2019-04-18
Field Backups was added to StorageDetails struct

karvounis:
Hello there,

I recently noticed that there is no Backups fields in the StorageDetails{} and that it was a TODO that was never done :) . We require it in my company to be able to have the array of backups UUIDs when we are requesting information regarding the details of a storage device. 
What is done in these commits:
- Added BackupUUIDs as an array of strings (follows the convention naming set by ServerUUIDs)
- Added a constant for the normal storage type
- Rearranged the StorageType constants by alphabetical order
- Added a test to unmarshal a StorageDetails struct

### DIFF
--- a/upcloud/storage.go
+++ b/upcloud/storage.go
@@ -4,10 +4,11 @@ import "encoding/xml"
 
 // Constants
 const (
-	StorageTypeDisk     = "disk"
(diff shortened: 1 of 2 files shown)


cancerballs:
Thanks for the PR!

PR #47, opened 2019-12-01
Implemented https://developers.upcloud.com/1.3/13-networks/

DGollings:
By the time I figured out it wasn't going to solve the problem I was trying to solve in the first place I was almost done implementing all the calls. So hope someone finds this useful

There does appear to be a bug in the Upcloud API though, no matter what I tried I would get an error concerning invalid values for ip_network(s)
https://developers.upcloud.com/1.3/13-networks/#create-sdn-private-network

This included running the exact create sdn private network command, request and saving the output to xml and sending _that_ back. Same error.

So switched to JSON for that bit (it worked), but I assume it would be better to fix it properly at Upcloud (or in my code). Just look for FIXME and you'll find the relevant bits to remove once it works

### DIFF
--- a/upcloud/client/client.go
+++ b/upcloud/client/client.go
@@ -97,7 +97,6 @@ func (c *Client) PerformPostRequestJsontoXMLFIXME(url string, requestBody []byte
 	if requestBody != nil {
 		bodyReader = bytes.NewBuffer(requestBody)
 	}
-	fmt.Printf("%s", requestBody)
 
 	request, err := http.NewRequest("POST", url, bodyReader)
 
--- a/upcloud/client/client.go
+++ b/upcloud/client/client.go
@@ -14,7 +14,7 @@ import (
 
 // Constants
 const (
-	DEFAULT_API_VERSION = "1.2.3"
+	DEFAULT_API_VERSION = "1.3.0"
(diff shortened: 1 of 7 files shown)


Jalle19:
You should report the JSON vs. XML issue to UpCloud. The only reason this SDK uses XML in the first place is because the JSON structure is so horrible it's hard to marshal/unmarshal easily in Go.

DGollings:
@Jalle19 I might be crazy, but considering the repo name I figured I was speaking to Upcloud employees :)

Just report via the upcloud chat thing?

Jalle19:
Well, depends if any of them are actively looking at these issues I guess :-P

Personally I'd send an e-mail to support@upcloud.com, they usually answer pretty quickly.

DGollings:
@Jalle19 they've confirmed its a bug, hopefully it can be fixed soon so the pull request can be pulled in more cleanly

PopoSensei:
Hello from 2020. Unfortunately we have updated the go-api so that this PR is unneeded. Therefore I will close this PR without merging. See to latest changes in [ChangeLog](https://github.com/UpCloudLtd/upcloud-go-api/blob/master/CHANGELOG.md) for further details.

PR #49, opened 2020-07-17
Extends account, adds account list and account details to api

samdixon:
Hello UpCloud team,

I noticed a couple features were missing from the account api in the golang library. Below is a summary of the things I either extended or added. Let me know if there are any changes that need to be made before merging. 

- Extends account struct and service by adding Resource Limits which is a default response item from the '/account' api.
- Adds struct and service for '/account/list' api.
- Adds struct and service for '/account/detail/username' api.
- Adds and updates tests for new items and extensions.

### DIFF
--- a/upcloud/account.go
+++ b/upcloud/account.go
@@ -2,6 +2,66 @@ package upcloud
 
 // Account represents an account
 type Account struct {
-	Credits  float64 `xml:"credits"`
-	UserName string  `xml:"username"`
+	Credits        float64        `xml:"credits"`
+	UserName       string         `xml:"username"`
+	ResourceLimits resourceLimits `xml:"resource_limits"`
+}
+
+type resourceLimits struct {
+	Cores      int `xml:"cores"`
+	Memory     int `xml:"memory"`
+	Networks   int `xml:"networks"`
+	PublicIpv4 int `xml:"public_ipv4"`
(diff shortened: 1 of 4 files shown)


upcloud/account.go
```diff
@@ -2,6 +2,66 @@ package upcloud
...
+	UserName               string   `xml:"username"`
+	VatNumber              string   `xml:"vat_number"`
+	NetworkAccess          []string `xml:"network_access>network"`
+	ServerAccess           []server `xml:"server_access>server"`
+	StorageAccess          []string `xml:"storage_access>storage"`
+	TagAccess              []tag    `xml:"tag_access>tag"`
+	IpFilter               []string `xml:"ip_filters>ip_filter"`
+}
+
+type server struct {
```
Jalle19, line 59: What are these for?

PopoSensei:
Unsure if there really is a need for account list in the API. Also we have recently updated the whole repo to JSON and UpCloud API 1.3. Also for what purpose would there be to get account details through this API.

Personally I would prefer users administrate their accounts through the UpCloud hub. However if there is a reasonable described need for any of these in this API please let us know.

I will close this PR, without merging, next week, if no changes are implemented and reasoning provided.

samdixon:
I will go ahead and close

PR #51, opened 2020-08-03
Convert initial 3 functions to JSON API

yoink00:
Changes:

* Switches GetServers, CreateServer and DeleteServerAndStorages to use JSON API as opposed to the XML API.
* Adds a simple example use of the API
* Adds an Error unmarshal test


upcloud/service/service.go
```diff
@@ -110,13 +111,17 @@ func (s *Service) GetServerConfigurations() (*upcloud.ServerConfigurations, erro
...
-	response, err := s.basicGetRequest("/server")
+	response, err := s.basicJSONGetRequest("/server")
 
 	if err != nil {
-		return nil, parseServiceError(err)
+		return nil, parseJSONServiceError(err)
 	}
 
-	xml.Unmarshal(response, &servers)
+	fmt.Println(string(response))
```
PopoSensei, line 120: Was this left from testing?

yoink00:
Yes. Let me fix that.

examples/upcloud-cli/main.go
```diff
@@ -0,0 +1,202 @@
...
+		StorageDevices: []upcloud.CreateServerStorageDevice{
+			{
+				Action:  upcloud.CreateServerStorageDeviceActionClone,
+				Storage: "01000000-0000-4000-8000-000050010400",
+				Title:   "Centos8 from a template",
+				Size:    50,
+				Tier:    upcloud.StorageTierMaxIOPS,
+			},
+		},
+		IPAddresses: []request.CreateServerIPAddress{
```
aakso, line 90: It would be better not to use IPAddresses field anymore in examples as it's functionality has been replaced by the Networking structure.

See https://developers.upcloud.com/1.3/8-servers/#create-server

yoink00: Hi Anton,

At the moment I am in the process of converting the 1.2 API from XML to JSON before moving on to using the 1.3 API. This example is correct as of 1.2? I will update it once I begin migrating from 1.2 to 1.3.

Thanks,

Stuart

aakso: Hey Stuart,

(Apparently github ate my reply)

Anyway, the field is valid for 1.2. However, as it can't express the newer network features the usage is not encouraged.

Thanks,
Anton

aakso:
General comment about the API Client:
I'd like to see either a setter and a getter for the `*http.Client` or have it as a public field. Would be more developer friendly.

yoink00:
> General comment about the API Client:
> I'd like to see either a setter and a getter for the `*http.Client` or have it as a public field. Would be more developer friendly.

I thought the same thing. PR #53 adds `go-vcr` to the service tests and as a side effect allows you to specify an `http.Client`.

PopoSensei:
LGTM

PR #56, opened 2020-08-05
XML-to-JSON conversion

yoink00:
Changes:
 * Converts all API calls to JSON (using 1.2 API)
 * Removes all XML tags and custom marshal/unmarshal
 * Adds custom JSON marshal/unmarshal functions where needed


upcloud/service/fixtures/getaccount.yaml
```diff
@@ -0,0 +1,45 @@
...
+               "cores" : 100,
+               "detached_floating_ips" : null,
+               "memory" : 307200,
+               "networks" : 100,
+               "public_ipv4" : null,
+               "public_ipv6" : 100,
+               "storage_hdd" : 10240,
+               "storage_ssd" : 10240
+            },
+            "username" : "opencredo"
```
PopoSensei, line 29: This file and endpoint overall might be a bit too user specific.

PopoSensei: I think though that the username is the only user specific datapoint

PR #57, opened 2020-08-07
Update to JSON API version 1.3 

yoink00:
Changes:
 * Updates "no change" endpoints to 1.3
 * Updates existing but changed endpoints to 1.3:
    * Tags
    * Account
    * Server
    * IPAddress

There will be some tweaks once the "added" resources are complete.

### DIFF
--- a/examples/upcloud-cli/main.go
+++ b/examples/upcloud-cli/main.go
@@ -87,10 +87,16 @@ func createServer(s *service.Service) error {
 				Tier:    upcloud.StorageTierMaxIOPS,
 			},
 		},
-		IPAddresses: []request.CreateServerIPAddress{
-			{
-				Access: upcloud.IPAddressAccessPrivate,
-				Family: upcloud.IPAddressFamilyIPv4,
+		Networking: &request.CreateServerNetworking{
+			Interfaces: []request.CreateServerInterface{
+				{
+					IPAddresses: []request.CreateServerIPAddress{
+						{
+							Family: upcloud.IPAddressFamilyIPv4,
+						},
+					},
+					Type: upcloud.IPAddressAccessUtility,
+				},
 			},
 		},
 	})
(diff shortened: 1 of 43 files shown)


upcloud/service/service.go
```diff
@@ -707,7 +711,7 @@ func (s *Service) GetTags() (*upcloud.Tags, error) {
 }
 
 // Wrapper that performs a GET request to the specified location and returns the response or a service error
-func (s *Service) basicJSONGetRequest(location string) ([]byte, error) {
+func (s *Service) basicFutureGetRequest(location string) ([]byte, error) {
```
PopoSensei, line 714: Why future and not just basicGetRequest?

PopoSensei: And why change this at all?

examples/upcloud-cli/main.go
```diff
@@ -87,10 +87,16 @@ func createServer(s *service.Service) error {
...
-				Family: upcloud.IPAddressFamilyIPv4,
+		Networking: &request.CreateServerNetworking{
+			Interfaces: []request.CreateServerInterface{
+				{
+					IPAddresses: []request.CreateServerIPAddress{
+						{
+							Family: upcloud.IPAddressFamilyIPv4,
+						},
+					},
+					Type: upcloud.IPAddressAccessUtility,
```
aakso, line 98: I'd suggest not to use constants from 1.2 here as the interface types in 1.3 are different thing.

upcloud/service/service.go
yoink00: Hi,

This is a bit of a hangover from the way I did the port. `basicFutureGetRequest` was for the `1.3` API and `basicJSONGetRequest` was for the `1.2` API.

I removed all references of `future` in the code once I finished but I missed this one. I spotted it yesterday and it's already fixed in the next batch of changes.

Would you like me to fix it now or are you happy to wait for the next PR?

Thanks,

Stuart

examples/upcloud-cli/main.go
yoink00: Hi,

Yes, I've spotted that too. I've added appropriate networking constants in the next batch of changes and switched everything to those.

Would you like me to fix it now or are you happy to wait for the next PR?

Thanks,

Stuart

aakso: Ok just wanting to check :) Let's get back to this in next PR then.

upcloud/service/service.go
PopoSensei: It's ok, for the next PR. 👍

PR #60, opened 2020-08-20
Allow empty boolean values and add fields to create server networking

trentrosenbaum:
This PR updates the the upcloud-go-api to allow for empty boolean values.  Updates cover tests and fixture files.
Also included in this PR are extra fields to allow network interfaces to be created with the server.


PopoSensei:
LGTM

PR #66, opened 2020-10-13
client/service: better timeout handling

aakso:
* client: only set default timeout if the caller hasn't provieded a HTTP
            client with non-zero timeout.
  * service: restore timeout in server start and stop while allowing additional
             time for the API to respond with an error.


PopoSensei:
LGTM

Issue #65, opened 2020-10-02
CreateStorageImport should allow passing io.Reader when type is direct_uploud


aakso:
As opening a file for reading is such a trivial task in Go I feel this should be the default.

Rationale:
user of this library might want to implement a progress bar / status. He/she might also want to calculate a checksum to be compared with the value that UpCloud API returns.

Proposal:
add another public field to CreateStorageImportRequest of type io.Reader or use an internal field with public setter.

PopoSensei:
Issue resolved in #67

PR #67, opened 2020-10-13
Storage import ioreader

aakso:
* request: SourceLocation is now untyped to allow passing of io.Reader
             instead of just string. This gives caller more flexibility
             while still providing ease of use. A new ContentType field
             allows the caller to specify if the locally available image
             is compressed or not.
  * service: support io.Reader by type asserting the SourceLocation.
             If it's string then it's used as a file name. Otherwise the
             io.Reader is passed as-is. Content type is set according
             to the request.
  * client:  support setting and getting of content type. Use the Set
             method instead of Add to reset the specified headers.

### DIFF
--- a/upcloud/client/client.go
+++ b/upcloud/client/client.go
@@ -23,9 +23,10 @@ const (
 
 // Client represents an API client
 type Client struct {
-	userName   string
-	password   string
-	httpClient *http.Client
+	userName    string
+	password    string
+	httpClient  *http.Client
+	contentType string
 }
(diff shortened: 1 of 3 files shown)


aakso:
resolves #65

tohalla (approved):
LGTM. Out of the scope of this pr, but should implement a function for adding generic headers (authorization, user-agent etc.) separate from the `addJSONRequestHeaders` in near future.

PR #68, opened 2020-10-13
fix server and storage import related type encoding/decoding

aakso:
* request: set SSHKeys to omitempty as the field is optional and will
             otherwise get validated by the API.
  * upcloud: fix StorageImportDetails.Completed field decoding. Use an alias
             struct and separately parse the completed timestamp if it's not
             an empty string. Also expose SourceLocation which is provided
             by the API.


tohalla (approved):
LGTM

PR #70, opened 2020-11-10
fix: omitempty on CreateServerStorageDevice

tohalla:
size is optional (defaulting to the size of the original) when cloning a storage

terraform provider allows omitting size for storage when creating a new one with clone action (throwing an error, as not supported by the sdk).

### DIFF
--- a/upcloud/request/server.go
+++ b/upcloud/request/server.go
@@ -58,7 +58,7 @@ type CreateServerStorageDevice struct {
 	Storage string `json:"storage"`
 	Title   string `json:"title,omitempty"`
 	// Storage size in gigabytes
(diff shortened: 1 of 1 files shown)


PopoSensei:
What happens if user (maybe not TF user) tries to create a new storage without size?

tohalla:
CreateServerStorageDevice only used when creating storages within the server creation request. Returns error from the API if the operation is not permitted (i.e. when action != clone and storage size not set).

PopoSensei:
Ok. Then LGTM, if tests pass can merge.

aakso:
looking good. Most of the fields should be omitempty in the first place unless the zero value has some meaning.

PR #71, opened 2020-11-10
Object storage

iivlei:
This PR adds all CRUD operations for Object Storage.

API: https://developers.upcloud.com/1.3/15-object-storage/

### DIFF
--- a/upcloud/client/client.go
+++ b/upcloud/client/client.go
@@ -14,7 +14,7 @@ import (
 
 // Constants
 const (
-	DefaultAPIVersion = "1.3.4"
+	DefaultAPIVersion = "1.3.6"
 	DefaultAPIBaseURL = "https://api.upcloud.com"
 
 	// The default timeout (in seconds)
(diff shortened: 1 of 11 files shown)


upcloud-api.postman_collection.json
```diff
@@ -523,6 +523,41 @@
...
+						"com"
+					],
+					"path": [
+						"1.3",
+						"object-storage",
+						""
+					]
+				}
+			},
+			"response": []
```
aakso, line 560: I don't know if we should keep updating this postman config file. I feel it's really not part of this pkg.

upcloud/client/client.go
```diff
@@ -14,7 +14,7 @@ import (
 
 // Constants
 const (
-	DefaultAPIVersion = "1.3.4"
+	DefaultAPIVersion = "1.3.6"
```
PopoSensei, line 17: IMO: Feels weird to keep this up to date by hand. Though don't know how to implement this in any other way 😅

upcloud-api.postman_collection.json
iivlei: I second this. Also too easy to forget to update, since it's not mandatory for the feature itself to function properly. Should we remove that file?

PopoSensei, line 560: Would want the rest of the endpoints here too. So that all the CRUD methods can be tested through postman as well.

iivlei: @PopoSensei see conversation up above https://github.com/UpCloudLtd/upcloud-go-api/pull/71#discussion_r521933313

PopoSensei: Oh. If we decide to not support, then forget this :)

aakso: maybe delete the changes and create a second PR to remove it?

upcloud/client/client.go
aakso: This should be a variable that is set in release phase thru the linker options. Let's anyway address this in a separate PR

aakso:
Please squash your commits before merging this.

PopoSensei:
LGTM2

upcloud-api.postman_collection.json
Darep: https://github.com/UpCloudLtd/upcloud-go-api/pull/72

PR #74, opened 2020-11-18
refactor: do not set contenttype for client

tohalla:
- set different content-type on request level
- expose method for dispatching request object


aakso:
this is better solution than my previous one. LGTM!

aakso:
even though we are removing public methods there is no release that includes those so should be ok.

PR #75, opened 2020-11-19
user-agent

tohalla:
feature: expose UserAgent -field which will be passed as a header for requests

### DIFF
--- a/internal/globals.go
+++ b/internal/globals.go
@@ -1,0 +1,3 @@
+package globals
+
+var Version = "3.0.0"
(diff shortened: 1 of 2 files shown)


ajmyyra (commented):
Could we also have the current version number here?

tohalla:
> Could we also have the current version number here?

Thought about it as well, but would like to have some automation for the version number management. Left out for now as it is likely that will forget to update the number with new releases

aakso:
maybe we should include the CI job in this PR too? Manual updating will lead to problems.

PR #80, opened 2020-12-03
fix: add omitempty for fields in account resourcelimits

tohalla:
unlimited resources were displayed as zero values (0) which was confusing


PopoSensei:
LGTM

PR #81, opened 2020-12-17
fix: backup_rule deletion

tohalla:
the only way to delete the backup rule is to pass backup_rule: {} to the modify storage request. Currently this is not possible with the go api


aakso:
looks good!

PR #82, opened 2020-12-18
Migrate to GitHub actions

tohalla:


README.md
```diff
@@ -1,6 +1,6 @@
 # UpCloud Go API client library
 
-[![Build Status](https://travis-ci.org/UpCloudLtd/upcloud-go-api.svg?branch=master)](https://travis-ci.org/UpCloudLtd/upcloud-go-api)
+![Build Status](https://github.com//UpCloudLtd/upcloud-go-api/workflows/Upcloud%20go%20api%20test/badge.svg)
```
PopoSensei, line 3: I'm guessing this is a typo `https://github.com//UpCloudLtd`and should be with one /

PopoSensei: Also this change is missing from the Python api. So a good catch 😄

PopoSensei:
Other than the badge typo LGTM 🌮

github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
...
+    branches:
+      - master
+jobs:
+  Test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go-version: [ 1.14.x, 1.15.x ]
+    steps:
+      - name: Test
```
PopoSensei, line 16: Oh yeah. This steps name should probably be `setup` or `install` or `init env`. Since that is what the step is doing

tohalla: Did some changes (some cleaning up), any thoughts?

tohalla: could/should probably rename the build to test (also the script), but left like this for now

PopoSensei: naming steps is IMO a good practice. At least when the step is doing complicated or unusual stuff. This seems ok to me though :)

github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
+name: Upcloud go api test
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
PopoSensei, line 8: Hmmm... Unsure if the CI/CD runs against PR:s, since at least this PR is not checking anything. My TF one is running the GitHub actions.

tohalla: resolved by fixing the path 😁

PR #85, opened 2021-01-21
feature: add missing index field

tohalla:


tohalla:
api does not seem to support this

PR #88, opened 2021-05-20
Test: fix -short in tests

moitias:
It seems like some tests that should be skipped with -short weren't, which is fixed with this PR.

This includes *all* tests using record(), as they all seem to require credentials to be set. (?)

### DIFF
--- a/upcloud/service/service_test.go
+++ b/upcloud/service/service_test.go
@@ -319,9 +319,6 @@ func TestErrorHandling(t *testing.T) {
 // - stops the server
 // - deletes the server
 func TestCreateModifyDeleteServer(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping test in short mode")
-	}
 	t.Parallel()
 
 	record(t, "createmodifydeleteserver", func(t *testing.T, svc *Service) {
@@ -398,9 +395,6 @@ func TestCreateModifyDeleteServer(t *testing.T) {
 // - creates a server
 // - deletes the server including storage
 func TestCreateDeleteServerAndStorage(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping test in short mode")
-	}
 	t.Parallel()
 
 	record(t, "createdeleteserverandstorage", func(t *testing.T, svc *Service) {
@@ -494,9 +488,6 @@ func TestGetIPAddresses(t *testing.T) {
 // - modifies the PTR record of the IP address
 // - deletes the IP address
 func TestAttachModifyReleaseIPAddress(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping test in short mode")
-	}
 	t.Parallel()
 
 	record(t, "attachmodifyreleaseipaddress", func(t *testing.T, svc *Service) {
@@ -671,9 +662,6 @@ func TestAttachModifyReleaseFloatingIPAddress(t *testing.T) {
 // - deletes the firewall rule
 //
 func TestFirewallRules(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping test in short mode")
-	}
 	t.Parallel()
 
 	record(t, "firewallrules", func(t *testing.T, svc *Service) {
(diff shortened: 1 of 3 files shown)


aakso:
If you are skipping the tests that call record are they still ran in short mode using the saved fixtures?

moitias:
> If you are skipping the tests that call record are they still ran in short mode using the saved fixtures?

record() calls getCredentials() which panics without credentials set, which - to me - seems to be what -short does. From the readme;

> You can skip running the integration tests and just run the unit tests by passing -short to the test command.

which doesn't *exactly* say that it's for running tests without the credentials, but record() tests would fall under integration tests, no?

aakso:
> which doesn't _exactly_ say that it's for running tests without the credentials, but record() tests would fall under integration tests, no?

getCredentials() doesn't panic if UPCLOUD_GO_SDK_TEST_NO_CREDENTIALS is enabled. Forgive me if I don't understand the codebase completely but it's been my understanding that part of the deal with go-vcr is that we get fixtures we can test later againt verifying that our local parsing and logic works.

moitias:
> > which doesn't _exactly_ say that it's for running tests without the credentials, but record() tests would fall under integration tests, no?
> 
> getCredentials() doesn't panic if UPCLOUD_GO_SDK_TEST_NO_CREDENTIALS is enabled. Forgive me if I don't understand the codebase completely but it's been my understanding that part of the deal with go-vcr is that we get fixtures we can test later against verifying that our local parsing and logic works.

Sure. However;

1. What this PR does is essentially expand -short to all cases similar to where -short is currently used. Tests that do currently skip with -short all are of the record()-type and skip it right away without any concern for the environment variable. I also can't find any pattern to why these particular tests should be skipped with -short; they are TestCreateModifyDeleteStorage, TestAttachDetachStorage, TestCloneStorage, TestTemplatizeServerStorage, TestCreateModifyDeleteServer, TestCreateDeleteServerAndStorage, TestAttachModifyReleaseIPAddress, TestLoadEjectCDROM, TestCreateRestoreBackup, TestFirewallRules and TestTagging. I should and will remove the check from these if we go with the approach in the MR, of course.

2. I'm not sure what the use case for having a separate -short and UPCLOUD_GO_SDK_TEST_NO_CREDENTIALS is? Do we reallly need such granularity and why? Especially separately - I could understand it if one of them implied the other (eg. they would be sort of levels) but I don't really see when we would want to do only one or the other. It seems to complicate testing unnecessarily for me. One option here could be to say that -short does imply NO_CREDS, and change the fake credentials check in getCredentials() to include checking for -short - but then the question becomes what use is the environment variable?

3. The environment parameter is not documented at all. That is, it's impossible to even run tests by reading documentation but without delving into the code, which doesnt seem to be the greatest situation to be in. My view is that we should make the barriers for testing as low as possible in order to make people actually run them rather than get frustrated over trying to figure out what these switches imply and how do they differ. This could of course be improved by improving the documentation, but to do that we would actually need to figure out the difference between -short and NO_CREDS which is still very unclear to me.

Based on these, my suggestion would be removing the NO_CREDS flag and not calling getCredentials if -short is set, basically moving setting the faked username/password from getCredentials() to record(), if -short is set rather than only if NO_CREDS is set. To be clear, this would mean that the recorded tests would be ran with -short with faked credentials and environment credentials without.

Also; I do kind of feel that having integration tests (that actually create resources that cost money) to be the default when running `go test ./...` to be a bit -.. backwards. Default test, IMO, should be something that does succeed without any special flags or environment variables needing to be set. If I was making something like this from scracth, I would personally go with go test *not* running the actual integration tests and put the integration tests behind a flag / build tag. However, I'm fine with just dropping NO_CREDS and documenting -short clearly and moving on for now.

moitias:
Also, it really does seem to me that at least the original intention for -short was indeed to skip tests such as these, as the documentation clearly states that -short allows one to run unit tests only, skipping integration tests and the go-vcr/record() style tests definitely test integration.

moitias:
Also, turns out that the CI pipeline in GH actions runs the tests with NO_CREDS, which means that the actual integration tests against a real backend are not run except by manually by individual developers..

kaminek:
LGTM

PR #89, opened 2021-05-21
chore: remove ad-hoc build.sh

moitias:
- remove ad-hoc build.sh (which did not build, but ran some linters and tests) and related scripts 
- run the same (and some additional) linters with golangci-lint instead and add golangci-lint to precommit
- update gh actions to do the linting and running tests in separate steps


.github/workflows/main.yml
```diff
@@ -13,16 +13,20 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [ 1.14.x, 1.15.x ]
+        go-version: [ 1.14.x, 1.15.x, 1.16.x ]
```
kaminek, line 16: do we really need to support older golang version?

kaminek (approved):
LGTM

.github/workflows/main.yml
moitias: Mmmh, it's a good question - maybe for an SDK it doesn't hurt to test against older versions too, as our users might be on older versions and averse to updating? 

Although, the tests here should _probably_ run the full test suite, including the actual backend integration tests that are now disabled, which can take a while so adding more versions to test might make the tests even more lengthy as - from what I understand - all the tests share creds so we cant even run them in parallel..

moitias: Yeah, after a little think I think we should keep these. This is because I feel we should be quite bconservative when using any brand new language features in the SDK in order to support users who can't or won't update their Go versions.  Thus, the older versions serve the purpose of making sure that we're not using any such features. 

The question this raises, however, is whether 1.14,.1.15 and 1.16 is the selction of versions we actually do want to test again? Would it make more sense to test, for example, 1.10 and 1.16 only - eg. figure out the minimum required version (1.10 picked randomly for this example) and test against that and the very latest version?

PR #93, opened 2021-06-09
feat: add req/res dump for debug

kaminek:
Signed-off-by: Amine Kherbouche <kaminek92@gmail.com>


kaminek:
maybe missing doc update

moitias:
not sure if we want to commit to a logging packge - especially one that is not the same that we use elsewhere
it's already possible for the consumer to intercept the requests - by passing in a custom http.client
i'm not sure about the dumping itself, seems like a custom format and api really shouldnt concern itself with formatting things and much less the terminal
also not a big fan of dumpHTTPRequest modifying the request itself, sort of behind the scenes

PR #95, opened 2021-07-06
client: custom api uri and version from env

kaminek:
Signed-off-by: Amine Kherbouche <kaminek92@gmail.com>

### DIFF
--- a/upcloud/client/client.go
+++ b/upcloud/client/client.go
@@ -9,6 +9,8 @@ import (
 	"time"
 
 	globals "github.com/UpCloudLtd/upcloud-go-api/internal"
+	upcloud "github.com/UpCloudLtd/upcloud-go-api/upcloud"
+
 	"github.com/blang/semver"
 	"github.com/hashicorp/go-cleanhttp"
 )
@@ -18,6 +20,8 @@ const (
 	DefaultAPIVersion = "1.3.6"
 	DefaultAPIBaseURL = "https://api.upcloud.com"
 
+	APIVersionEnv = "UPCLOUD_API_VERSION"
+	APIBaseURLEnv = "UPCLOUD_API_URL"
 	// The default timeout (in seconds)
 	DefaultTimeout = 60
 )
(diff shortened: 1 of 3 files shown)


moitias:
This is an awesome improvement (which also will allow us to use upctl against local dev.. finally!) but I'm not 100% sure on whether taking the url from environment is that great of an idea for a library? It kind of obscures the source of the URL and seems like it could result in hard to debug situations.

I'd much rather prefer the initialization methods / the client struct itself have a configurable endpoint URL. Adding it to the initialization method will break backwards compatibility, of course, but if that's crucial, make a separate intialization method for that case and/or add a SetBaseURL() method?

upcloud/client/client.go
```diff
@@ -33,7 +40,18 @@ type Client struct {
 
 // New creates ands returns a new client configured with the specified user and password
 func New(userName, password string) *Client {
-	return NewWithHTTPClient(userName, password, cleanhttp.DefaultClient())
+	httpClient := &http.Client{}
+	httpTransport := cleanhttp.DefaultTransport()
+
+	if upcloud.IsEnvEnabled(APIUnsecureSkipVerifyEnv) {
```
moitias, line 46: Do note that NewWithHTTPClient should also support the same semantics. 

Although, I'm not sure how it would / should work if UPCLOUD_API_INSECURE_SKIP_VERIFY = true and the custom client does not include that..

upcloud/client/client.go
```diff
@@ -184,9 +203,12 @@ func (c *Client) PerformRequest(request *http.Request) ([]byte, error) {
 
 // Returns the base URL to use for API requests
 func (c *Client) getBaseURL() string {
-	urlVersion, _ := semver.Make(DefaultAPIVersion)
+	url := upcloud.GetEnvOrDefault(APIBaseURLEnv, DefaultAPIBaseURL)
+	version := upcloud.GetEnvOrDefault(APIVersionEnv, DefaultAPIVersion)
```
moitias, line 207: not sure if we really need two different environment variables to control one thing (eg. the endpoint url?) could we not just pass in the version in the url if we're overriding the base url anyhow?

peknur:
Closing this PR as duplicate.  Issue is mostly resolved in #148 and #129. If we need custom versioning also let's create new PR.

PR #96, opened 2021-08-27
feat: implement managed database support

aakso:
* upcloud, request, service: add necessary types and methods for managed
  databases.
* test: change record-helper to pass *recorder.Recorder to the callback.
  This allows executing specific test code during recording vs playback.
* test: add UPCLOUD_GO_SDK_API_HOST envvar to allow running integration
  tests against another API host.
* go.mod: add mysql and postgres drivers for testing only.


upcloud/client/client.go
```diff
@@ -63,6 +68,14 @@ func (c *Client) GetTimeout() time.Duration {
 	return c.httpClient.Timeout
 }
 
+// WithCustomAPI returns a clone of the client with overridden baseUrl and version
+func (c *Client) WithCustomAPI(baseUrl, version string) *Client {
```
moitias, line 72: Minor question: Is there a reason for doing it this way and not just exposing the variables, having Set-methods (like `SetTimeout` that's already there) or allowing them to be used like other configuration - eg. in the initialization methods? 

Adding more permutations to the init methods isn't ideal in any way but tbh I'm not sure why the extra layer of abstraction is a good idea either and this adds something extra to grok / a potential (but probably unlikely) footgun with the cloning (while keeping the same reference to http.Client)?

upcloud/client/client.go
```diff
@@ -184,9 +209,9 @@ func (c *Client) PerformRequest(request *http.Request) ([]byte, error) {
 
 // Returns the base URL to use for API requests
 func (c *Client) getBaseURL() string {
-	urlVersion, _ := semver.Make(DefaultAPIVersion)
+	urlVersion, _ := semver.Make(c.apiVersion)
```
moitias, line 212: Should we check for errors here, in case the user has entered an invalid version? Panic should be fine, I guess, unless you want to add error handling here as well? Otherwise, it might be kind of unclear as `urlVersion` would (probably?) end up as "", and the error message then being kind of odd with just a faulty URL with no other indications of what went wrong?

upcloud/managed_database.go
```diff
@@ -0,0 +1,693 @@
...
+	// native API.
+	ManagedDatabaseUserTypeNormal = "normal"
+)
+
+// Authentication types
+const (
+	// ManagedDatabaseUserAuthenticationCachingSHA2Password selects "caching_sha2_password" type of authentication type.
+	// This type is only supported with MySQL services.
+	//nolint:gosec
+	ManagedDatabaseUserAuthenticationCachingSHA2Password = "caching_sha2_password"
```
moitias, line 81: I'm guessing the gosec warning is related to it thinking this is a password due to the name? It would be good to add a nolint comment clarifying it, eg something like `//nolint:gosec // this is not actually a password but authentication type`

upcloud/managed_database.go
```diff
@@ -0,0 +1,693 @@
...
+	Type             string                          `json:"type"`
+	UpdateTime       time.Time                       `json:"update_time"`
+	ServiceURI       string                          `json:"service_uri"`
+	ServiceURIParams ManagedDatabaseServiceURIParams `json:"service_uri_params"`
+	Users            []ManagedDatabaseUser           `json:"users"`
+	UUID             string                          `json:"uuid"`
+	Zone             string                          `json:"zone"`
+}
+
+// ManagedDatabaseBackup represents a taken full backup's point in time
```
moitias, line 111: This comment could use some editing; `.. represents a full backup taken at a point in time` or something?

upcloud/managed_database.go
```diff
@@ -0,0 +1,693 @@
...
+	Host      string `json:"host"`
+	Port      int    `json:"port"`
+	// Route describes how the component can be reached.
+	// 	public  = The component can be reached over public internet
+	// 	dynamic = The component can be only reached over utility network
+	Route string `json:"route"`
+	// Usage describes the role of the component.
+	// 	primary = The component is a primary (writable) instance in a cluster
+	// r	eplica = The component is a standby (read-only) instance in a cluster
+	Usage string `json:"usage"`
```
moitias, line 129: Should these be made into consts / custom types / Go-style enums, if these are the only options (currently) available?

upcloud/managed_database.go
```diff
@@ -0,0 +1,693 @@
+package upcloud
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
+// Managed Database states
+const (
```
moitias, line 10: Is there a reason to keep all of these as strings and not as custom enum(ish) types?

upcloud/managed_database.go
```diff
@@ -0,0 +1,693 @@
...
+	if err := json.Unmarshal(d, &in); err != nil {
+		return err
+	}
+	*m = ManagedDatabaseConnection(in.alias)
+	m.QueryDuration = time.Duration(in.QueryDuration * float64(time.Second))
+	return nil
+}
+
+// ManagedDatabaseMaintenance represents the set time of week when automatic maintenance operations are allowed
+type ManagedDatabaseMaintenance struct {
```
moitias, line 186: Could make the relation to time explicit in the name of the type (`ManagedDatabaseMaintenanceTime`?), but not a big deal.

upcloud/managed_database.go
```diff
@@ -0,0 +1,693 @@
...
+	// Rows contains a slice of values per row. The inner slice has the same indexing as the
+	// ManagedDatabaseMetricsChartHeader.Columns
+	Rows [][]float64
+}
+
+// MarshalJSON implements json.Marshaler for the canonical form of metrics chart
+func (m ManagedDatabaseMetricsChartFloat64) MarshalJSON() ([]byte, error) {
+	var chart canonicalManagedDatabaseMetricChartForMarshal
+	chart.Data.Columns = make([]ManagedDatabaseMetricsColumn, 0, len(m.Columns)+1)
+	chart.Data.Columns = []ManagedDatabaseMetricsColumn{{Label: "time", Type: "date"}}
```
moitias, line 221: Doesn't this re-create the slice, eg. the previous line with make is a no-op? Eg. should this really be `chart.Data.Columns = append(chart.Data.Columns, { Label: "time", Type: "date" })`?

upcloud/managed_database.go
```diff
@@ -0,0 +1,693 @@
...
+	// Rows contains a slice of values per row. The inner slice has the same indexing as the
+	// ManagedDatabaseMetricsChartHeader.Columns
+	Rows [][]int
+}
+
+// MarshalJSON implements json.Marshaler for the canonical form of metrics chart
+func (m ManagedDatabaseMetricsChartInt) MarshalJSON() ([]byte, error) {
+	var chart canonicalManagedDatabaseMetricChartForMarshal
+	chart.Data.Columns = make([]ManagedDatabaseMetricsColumn, 0, len(m.Columns)+1)
+	chart.Data.Columns = []ManagedDatabaseMetricsColumn{{Label: "time", Type: "date"}}
```
moitias, line 322: Same as above, should probably be append() rather than creating a new slice?

upcloud/managed_database.go
```diff
@@ -0,0 +1,693 @@
...
+	if m == nil {
+		return nil
+	}
+	if *m == nil {
+		*m = make(ManagedDatabaseProperties)
+	}
+	return (*m)[name]
+}
+
+// GetString returns a string property value. An empty string is returned for non-existing or not-a-string values.
```
moitias, line 414: I'm a bit iffy on the 'return default value  on wrong-type values' approach, seems like it could lead to errors that are very easy to miss. Would returning an error in that case make the other bits too unwieldy?

upcloud/managed_database.go
```diff
@@ -0,0 +1,693 @@
...
+	// Offset describes the next available offset. Use this to query more logs.
+	Offset string `json:"offset"`
+	Logs   []ManagedDatabaseLogEntry
+}
+
+type ManagedDatabaseLogEntry struct {
+	Hostname string    `json:"hostname"`
+	Message  string    `json:"msg"`
+	Time     time.Time `json:"time"`
+	Unit     string    `json:"unit"`
```
moitias, line 484: Unit could maybe use a comment? It's not clear to me (at least) what a unit refers to in the context of a log entry.

upcloud/managed_database.go
```diff
@@ -0,0 +1,693 @@
...
+		Quantile99:           uint64(m.Quantile99) * 1e3,
+		Quantile999:          uint64(m.Quantile999) * 1e3,
+		QuerySampleTimerWait: uint64(m.QuerySampleTimerWait) * 1e3,
+		SumTimerWait:         uint64(m.SumTimerWait) * 1e3,
+	}
+	return json.Marshal(&out)
+}
+
+func (m *ManagedDatabaseQueryStatisticsMysql) UnmarshalJSON(d []byte) error {
+	// Duration values are in picoseconds in the canonical form
```
moitias, line 588: Seems a bit odd that the durations are different across pg (float64 milliseconds?) and mysql (uint64 picoseconds) - could we make them consistent?

upcloud/request/managed_database.go
```diff
@@ -0,0 +1,549 @@
...
+type CancelManagedDatabaseConnection struct {
+	// UUID selects a managed database instance to manage queries of
+	UUID string
+	// Pid selects a connection pid to cancel or terminate
+	Pid int
+	// Terminate selects whether the connection will be forcefully terminated
+	Terminate bool
+}
+
+// RequestURL implements the Request interface
```
moitias, line 26: Could document the package where `Request` is, as well?

upcloud/request/managed_database.go
```diff
@@ -0,0 +1,549 @@
...
+}
+
+// GetStringSlice returns a string-slice property value. A nil slice is returned for non-existing or not-a-string-slice values.
+// The underlying map is initialised if it's nil.
+func (m *ManagedDatabasePropertiesRequest) GetStringSlice(name string) []string {
+	if slice, ok := m.Get(name).([]string); ok {
+		return slice
+	}
+	// The slice might be a slice of empty interfaces if upcloud.ManagedDatabaseProperties is converted into
+	// request.ManagedDatabasePropertiesRequest
```
moitias, line 330: These two containers seem very close to each other and duplicate some code, could we DRY them up or if they do have some differences making it impossible, making the differences clear?

upcloud/request/managed_database.go
```diff
@@ -0,0 +1,549 @@
...
+	}
+	if *m == nil {
+		*m = make(ManagedDatabasePropertiesRequest)
+	}
+	(*m)[name] = value
+	return m
+}
+
+// SetString associates key with a string value. The underlying map is initialised if it's nil
+func (m *ManagedDatabasePropertiesRequest) SetString(name string, value string) *ManagedDatabasePropertiesRequest {
```
moitias, line 251: I'm guessing these are just for making the consumer code more self-documenting? Should we make `Set` private in that case, to enforce their usage?

moitias (approved):
Overall a very very nice PR, liking the extensive documentation a lot! Various smallish nitpicky comments but nothing serious.

A few high-level observations;
- could have split this to smaller pieces, the go-vcr updates and api url are not realllly related
- testing seems to be oriented towards correct semantics, but personally I would like to see some more testing of error handling / incorrect semantics as well.

but either of these isn't a big deal either so LGTM :+1:

upcloud/managed_database.go
moitias: Doing enums would also make these maybe a little less verbose and hopefully bit easier to understand and use? eg. `StateRunning ManagedDatabaseState = "running"`? (or even just `Running ManagedDatabaseState = "running"`)

upcloud/client/client.go
aakso: We can do this with Set-methods obviously. This was kind of a precursor for possible upcoming context support where mutating the client state is certainly undesired. We have basically two options there: either introduce new duplicate methods which will lead to potentially long method names `CancelManagedDatabaseConnection` -> `CancelManagedDatabaseConnectionContext` or `CtxCancelManagedDatabaseConnection` while other option being the stdlib http request's `Context` method. Personally I find `srv.WithContext(myCtx).CancelManagedDatabaseConnection(...)` easier to read and use. But I digress.

The api base url override was mainly added while developing this and can be certainly removed from this PR as there's now an envvar based override method for integration testing only. However, we will eventually need this for upctl development.

The With-stuff is purely my personal preference as I'd like an library to be thread-safe from the get go. However, we can change it to mutating method if you prefer it.

upcloud/client/client.go
aakso: yes, an oversight.

upcloud/managed_database.go
aakso: fair point, will add.

upcloud/managed_database.go
aakso: whoops :)

upcloud/managed_database.go
aakso: yes, though they are never used in requests so I'll get warnings of unused exported constants :)

upcloud/managed_database.go
aakso: I would certainly like to do that. However, as everything is in upcloud -package, wouldn't that be too generic name for an exported variable? We will certainly later have other constants that would like to be called `StateRunning`.

upcloud/managed_database.go
aakso: whoops :)

upcloud/managed_database.go
aakso: yes, will rename

upcloud/managed_database.go
aakso: I took inspiration from Vipers's GetString method. Obviously there's more complex semantics there of overlapping set of maps. Here we could just leave this as plain map[string]interface{}. Then again it would make working with slices more difficult for a user (for example the IPFilter property).

What do you think?

upcloud/managed_database.go
aakso: I think I will open a discussion on changing this in the API to `service`. Unit is certainly very misleading.

upcloud/managed_database.go
aakso: These are feeded directly from the database engine's statistics tables. I agree with you that we could represent these in unified unit but will need to agree with the team as there's already UI work being done. Should we just represent these as float64 in seconds or would it loose too much precision? Or just return them in nanoseconds so we can directly load them as `time.Duration`?

upcloud/request/managed_database.go
aakso: I'm a bit unfamiliar with the overall idea of this project's module layout. If you are ok with referencing upcloud pkg from request then yes. In fact, it seems I already do this.. :)

upcloud/request/managed_database.go
aakso: yes, though you can always just mutate the map. There are also some properties that have inner map as values. We provide setters for well-known properties that are common to all engines.

scop:
General naming remark: we're somewhat inconsistent with PostgreSQL related naming, it's "postgres" in a bunch of places, and "pg" in others.

If I could fix this tiny bit of the world I think I'd personally go with "PostgreSQL" (case variations obviously as appropriate/required) everywhere because it's "correct" even though it's a mouthful.

But we can't fix it everywhere, and "pg" is a quite established and concise way to refer to it, so perhaps the practical improvement here would be to use that consistently instead of "postgres" here and there.

upcloud/client/client.go
moitias: Another option would be just to enforce context in all of those calls? This would, of course, make it backwards incompatible but would also force the consumer to think about the context. This also seems (to me) to be quite common nowadays in the Go world. 

However, the main concern I have here is consistency. Adding With here would make the client have 3 different ways of configuring it (separate New-methods, Sets, and With-cloning) which I hope we can agree isn't the greatest situation to be in. (As a sidenote, there's also the `New(...opts)`-style somewhere in another repo AFAIR, adding one more style of initialization in our codebase)

Personally it doesn't really matter which style we go for, as long as we keep it consistent. I do agree that this would be a useful feature, too, so not really into dropping it altogether over a question of which style should we go for. That is - let's just make these consistent (either Sets or With or New) and break the backwards compatibility now rather than have the different styles and wanting to break it later? 

Also, still not really sold on the cloning approach, it seems to have plenty of pitfalls that are not obvious and would require essentially manual work - What actually gets copied (eg. is this member an interface or struct or..)? What gets changed when we actually change things in the struct? Why does this `(Foo *s) ChangeSomething()`-method not actually change it in all cases? Why can't you write such methods? Are all the members also safe to be copied / used from multiple threads? Not that I really have a good solution for this either but just spitballing on the challenges ahead..

upcloud/managed_database.go
moitias: Well, I guess one could add some references in tests for them, giving IDEA (I'm guessing? Sometimes the warnings are not that great..) some ideas?

upcloud/managed_database.go
moitias: Ahh, right, yeah, can't really do that then, I was still a bit in -libs mode after reading through it lately..  So, no win for verbosity but types would be cool!

upcloud/managed_database.go
moitias: For Viper, the 'empty on invalid type' probably makes more sense than for us - for us, it kinda seems to me like it would always be more or less problematic (what if the types change in the input, without us knowing/noticing? Everything would still *seem* to work without a hitch but wouldn't)? I see your point about IPFilter being quite difficult to use if we drop these, maybe just make these return errors in case the type cast fails? 
  
Could even add `Must*`-style methods (eg. panic on error) as well, allowing the consumer to choose their required level of confidence and convenience (and hopefully making them think explicitly about the consequences!) in the types they are receiving?

upcloud/managed_database.go
moitias: Ah, right, makes sense now :D

upcloud/managed_database.go
moitias: Either option seems fine to me, as long as it's documented. 

I can't see float accuracy being a problem, considering most amount of the significant numbers for our use case should be after the point, but then again, that also points towards just using integers eg. nanoseconds as it seems very unlikely that we'll ever go over the limit of uin64 nanoseconds (eg. 1<<64-1 / 1e9 / 60 / 60 / 24 / 12 =~ 593 years). I guess we could do 'the highest precision we get' (eg. microseconds?) but in case we ever get something that provides higher accuracy, that would be kinda sucky.

So, I guess if I had a gun to my head and was forced to pick, I would go for integer nanoseconds but certainly wouldn't mind float seconds either.

upcloud/request/managed_database.go
moitias: Ahh, right, didn't notice that they are in different packages.. Up to you, in different packages I don't feel it's that crucial to DRY up.

upcloud/client/client.go
aakso: Can say only that I agree with everything. Maybe it would be still beneficial to move this to another PR as it's not right now strictly required?

Personally I'm all for `New(...opts)` style and we can drop the `With` -stuff.

upcloud/managed_database.go
aakso: I will look into these and maybe add some dummy validation in the integration tests.

upcloud/client/client.go
moitias: SGTM :+1:

upcloud/managed_database.go
aakso: I just checked with the team and they are at least happy with nanoseconds normalisation. From the sdk's PoV the limit is the signed int64 and it's plenty. At least it would remove the custom marshaling/unmarshaling.

upcloud/managed_database.go
aakso: string aliases ok or would insist in enums? Those increase boilerplate a bit.

upcloud/managed_database.go
aakso: Ok, good idea. Let's use errors with getters and consider adding must getters too.

The property system schema is somewhat dynamic but still they should be predictable. I would like to code generate them at some point but currently the situation is too much in flux for that.

aakso:
> Overall a very very nice PR, liking the extensive documentation a lot! Various smallish nitpicky comments but nothing serious.

Thanks for that and for the comments which help improve this a lot.

> 
>     * could have split this to smaller pieces, the go-vcr updates and api url are not realllly related

Yes, as discussed already, I'd lean on removing the api url stuff from this. Are you ok with the integration test envvar though as currently I can't generate the test fixtures without it?
 
>     * testing seems to be oriented towards correct semantics, but personally I would like to see some more testing of error handling / incorrect semantics as well.

Yes, the coverage for the service is about 75% which could be improved. The API is still in flux as mentioned before so I'd suggest revisiting the coverage issue when removing the `EXPERIMENTAL` markers from the methods. Would you be ok with that?

moitias:
Yes, both of these would be very much fine :+1:

upcloud/managed_database.go
moitias: I would prefer them to be typed, mostly for DX purposes (eg. discoverability, completion, etc) and to avoid typo-related bugs, missed (manual/custom) strings in (possible future) refactoring, etc.

upcloud/managed_database.go
moitias: Sounds good, code generation would get big :+1: :from me so hope we get there at some point!

upcloud/managed_database.go
aakso: yes but to clarify: you mean an int type with a custom TextMarshaler / TextUnmarshaler?

moitias: ohhhh right, sorry, misunderstood you, must be the hour... type aliased strings are exactly what I meant, my bad.

upcloud/client/client.go
moitias: Not to muddle the conversation further but just to add another datapoint slightly relevant to this - there's also the `NewObject(conf ObjectConfig)`-style in our codebase

aakso: removed the code from this PR

upcloud/managed_database.go
aakso: implemented

upcloud/managed_database.go
aakso: implemented

upcloud/managed_database.go
aakso: implemented

upcloud/managed_database.go
aakso: implemented

upcloud/managed_database.go
aakso: implemented

upcloud/managed_database.go
aakso: implemented

upcloud/managed_database.go
aakso: implemented

upcloud/managed_database.go
aakso: implemented

upcloud/managed_database.go
aakso: implemented

upcloud/managed_database.go
aakso: implemented

upcloud/request/managed_database.go
aakso: implemented

aakso:
> General naming remark: we're somewhat inconsistent with PostgreSQL related naming, it's "postgres" in a bunch of places, and "pg" in others.

Changed naming to follow the official names of said projects: `MySQL` and `PostgreSQL`

PR #101, opened 2022-01-12
feat(account): Subaccount management

peknur:
Implement subaccount CRUD operations and account listing.

### DIFF
--- a/upcloud/account.go
+++ b/upcloud/account.go
@@ -1,6 +1,12 @@
 package upcloud
 
-import "encoding/json"
+import (
+	"encoding/json"
+)
+
+const (
+	AccountTypeSubaccount string = "sub"
+)
 
 // Account represents an account
 type Account struct {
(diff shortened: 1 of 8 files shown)


upcloud/account.go
```diff
@@ -38,3 +44,138 @@ func (s *Account) UnmarshalJSON(b []byte) error {
...
+// AccountIPFilters represents IP address restrictions on API access; if set, allowed only from the specified ranges.
+// Ranges can be specified in CIDR format, ranges separated by a dash, or as single IP addresses
+type AccountIPFilters struct {
+	IPFilter []string `json:"ip_filter"`
+}
+
+// AccountDetails represents detailed information about an account
+type AccountDetails struct {
+	MainAccount string `json:"main_account"`
+	Type        string `json:"type"`
```
ka-myl, line 100: Would it make sense to make a separate enum for type? I think it's useful to narrow the types as much as possible so that the future developers don't need to wonder what values other then "sub" are allowed here

upcloud/service/service.go
```diff
@@ -94,23 +88,6 @@ func New(client *client.Client) *Service {
...
-		return nil, err
-	}
-
-	err = json.Unmarshal(response, &account)
-	if err != nil {
-		return nil, err
-	}
-
-	return &account, nil
-}
```
Darep, line 112: Hmm, is removing this a breaking change? 🤔  If we don't want to bump the library that much, we could also leave this in and maybe call the account service here

peknur: I just moved those account related things to account.go file. It's in same package so there shouldn't be any visible changes to users.

upcloud/service/account.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/UpCloudLtd/upcloud-go-api/v4/upcloud"
+	"github.com/UpCloudLtd/upcloud-go-api/v4/upcloud/request"
+)
+
+type Account interface {
+	GetAccountList() (upcloud.AccountList, error)
+	GetAccount() (*upcloud.Account, error)
+	GetAccountDetails(r *request.GetAccountDetailsRequest) (*upcloud.AccountDetails, error)
+	CreateSubaccount(r *request.CreateSubaccountRequest) (*upcloud.AccountDetails, error)
+	ModifySubaccount(r *request.ModifySubaccountRequest) error
```
ka-myl, line 15: Pity we can't add the updated subaccount to the return values here ;/

ka-myl (approved):
LGTM. One nitpick about account type, but feel free to ignore if you disagree :)

upcloud/account.go
peknur: Good point I'll change that

upcloud/service/service.go
Darep: Ok, cheers! I'm not super familiar with Go so I thought you could somehow import this service directly and call `service.GetAccount()` which I was guessing could break 👍

upcloud/service/account.go
peknur: @Darep what do you think? All the other `service.Modify*` functions return object and error.  We could change `ModifySubaccount` to return also `(*upcloud.AccountDetails, error)` but that would mean making two requests (POST and then GET) in `ModifySubaccount` . Personally I would change the interface and make two requests and also make interface consistent with other interfaces.

Darep: Let's make it consistent :blush: So, one more GET 😄

PR #103, opened 2022-01-24
feat(storage): resize partition and filesystem

peknur:
New service resizes the last partition of a storage and the ext3/ext4/XFS/NTFS filesystem on that partition if the partition does not extend to the end of the storage yet.

### DIFF
--- a/upcloud/service/storage.go
+++ b/upcloud/service/storage.go
@@ -448,7 +448,7 @@ func (s *Service) ResizeStorage(r *request.ResizeStorageRequest) (*upcloud.Stora
 
 	resizeBackup := upcloud.StorageResizeBackup{}
 	if err = json.Unmarshal(response, &resizeBackup); err != nil {
-		return &resizeBackup, nil
+		return &resizeBackup, err
 	}
 
 	return &resizeBackup, nil
--- a/upcloud/service/storage.go
+++ b/upcloud/service/storage.go
@@ -31,7 +31,7 @@ type Storage interface {
 	GetStorageImportDetails(r *request.GetStorageImportDetailsRequest) (*upcloud.StorageImportDetails, error)
 	WaitForStorageImportCompletion(r *request.WaitForStorageImportCompletionRequest) (*upcloud.StorageImportDetails, error)
 	DeleteStorage(*request.DeleteStorageRequest) error
-	ResizeStorage(r *request.ResizeStorageRequest) (*upcloud.StorageResizeBackup, error)
+	ResizeFilesystem(r *request.ResizeStorageRequest) (*upcloud.ResizeFilesystemBackup, error)
 }
 
 var _ Storage = (*Service)(nil)
@@ -430,7 +430,7 @@ func (s *Service) WaitForStorageImportCompletion(r *request.WaitForStorageImport
 	}
 }
 
-// ResizeStorage resizes the last partition of a storage and the ext3/ext4/XFS/NTFS filesystem
+// ResizeFilesystem resizes the last partition of a storage and the ext3/ext4/XFS/NTFS filesystem
 // on that partition if the partition does not extend to the end of the storage yet.
 //
 // Before the resize is attempted, a backup is taken from the storage. If the resize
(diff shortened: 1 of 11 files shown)


upcloud/service/storage_test.go
```diff
@@ -376,3 +376,63 @@ func TestDirectUploadStorageImport(t *testing.T) {
...
+
+	record(t, "resizestorage", func(t *testing.T, rec *recorder.Recorder, svc *Service) {
+		// start server
+		serverDetails, err := createMinimalServer(svc, "TestResizeStorage")
+		require.NoError(t, err)
+
+		// stop server
+		require.NoError(t, stopServer(svc, serverDetails.UUID))
+
+		// detach storage
```
Darep, line 399: Resize should be possible without detaching, too

peknur: I think that storage needs to be detached before modification (increase disk size). Resizing partition/fs can be done to attached storage (line 425).

upcloud/service/storage.go
```diff
@@ -428,3 +429,27 @@ func (s *Service) WaitForStorageImportCompletion(r *request.WaitForStorageImport
...
+func (s *Service) ResizeStorage(r *request.ResizeStorageRequest) (*upcloud.StorageResizeBackup, error) {
+	response, err := s.client.PerformJSONPostRequest(s.client.CreateRequestURL(r.RequestURL()), nil)
+
+	if err != nil {
+		return nil, parseJSONServiceError(err)
+	}
+
+	resizeBackup := upcloud.StorageResizeBackup{}
+	if err = json.Unmarshal(response, &resizeBackup); err != nil {
+		return &resizeBackup, nil
```
ka-myl, line 451: shouldn't this be `return &resizeBackup, err`? Or even `return nil, err`?

upcloud/service/storage.go
```diff
@@ -428,3 +429,27 @@ func (s *Service) WaitForStorageImportCompletion(r *request.WaitForStorageImport
...
+// ResizeStorage resizes the last partition of a storage and the ext3/ext4/XFS/NTFS filesystem
+// on that partition if the partition does not extend to the end of the storage yet.
+//
+// Before the resize is attempted, a backup is taken from the storage. If the resize
+// succeeds, backup details are returned. It is advisable to keep the backup until
+// you have ensured that everything works after the resize.
+//
+// If the resize fails, backup is used to restore the storage to the state where it
+// was before the resize. After that the backup is deleted automatically.
+func (s *Service) ResizeStorage(r *request.ResizeStorageRequest) (*upcloud.StorageResizeBackup, error) {
```
ka-myl, line 442: Perhaps we should name it `ResizeStoragePartition`? `ResizeStorage` alone suggests that this function takes care of all required storage modifications, while in reality user has to do a modify storage request (with size change) first.

peknur: I agree that the name is not most descriptive but I think that on the other hand it follows the rest of the naming convention because that URL is `/storage/<UUID>/resize` . I think that technically name should be something like `ResizeStoragePartitionAndFilesystem` but then it wouldn't map nicely to our API docs. For me it would make more sense to change the API URL first although I know that it might be easier said than done :)

Well that's my two cents if you and @Darep think that it is better to rename I can do it.

Darep: It is tricky… For anybody just browsing through the code or docs, `ResizeStorageFilesystem` as the name (or `ResizeStoragePartition` or both) would give the best hint as to what that method is all about. Having it as `ResizeStorage` might make some users call it without calling ModifyStorage with `size` first since they might think ResizeStorage does both. Though I feel like you could make the same mistake even if it was `ResizeStorageFilesystem` so then, in the end, the name might not matter :joy:

peknur: What if we add new (e.g. `ModifyStorageSize`) function that would do modify and resize?

Darep: Hmm, no strong feelings, but maybe we shouldn't go there as it would be one more thing to maintain, and things are more flexible with the separate methods 😊

ka-myl: I wouldn't mind adding that, but even then I would keep the `ResizeStorage(PartitionAndFilesystem)` function exposed to the public. It could be beneficial in some cases, like in TF provider.

As far as the name goes, I get the argument about the mapping to API, but tbh the api endpoint is kind of misleading here, and the point of SDK is to make it easier to use API. We can keep this name of course, but then I would add some additional explanation to the doc comment, to make it absolutely explicit that this function is for resizing the partition and filesystem after the storage itself has been resized.

peknur: Thanks for comments! Renamed this to ResizeStorageFilesystem as it is more descriptive name.

PR #110, opened 2022-03-09
Feat/load balancer type

peknur:
Initial load balancer model draft

### DIFF
--- a/upcloud/load_balancer.go
+++ b/upcloud/load_balancer.go
@@ -2,76 +2,240 @@ package upcloud
 
 import (
 	"time"
+)
 
-	"github.com/google/uuid"
+type LoadBalancerMode string
+type LoadBalancerMatcherType string
+type LoadBalancerActionType string
+type LoadBalancerStringMatcherMethod string
+type LoadBalancerHTTPMatcherMethod string
+type LoadBalancerIntegerMatcherMethod string
+type LoadBalancerMemberType string
+type LoadBalancerOperationalState string
+type LoadBalancerConfiguredStatus string
+
+const (
+	LoadBalancerModeHTTP LoadBalancerMode = "http"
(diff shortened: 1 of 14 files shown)


upcloud/load_balancer.go
```diff
@@ -4,72 +4,238 @@ import (
...
+
+	LoadBalancerHTTPMatcherMethodGet     LoadBalancerHTTPMatcherMethod = "GET"
+	LoadBalancerHTTPMatcherMethodHead    LoadBalancerHTTPMatcherMethod = "HEAD"
+	LoadBalancerHTTPMatcherMethodPost    LoadBalancerHTTPMatcherMethod = "POST"
+	LoadBalancerHTTPMatcherMethodPut     LoadBalancerHTTPMatcherMethod = "PUT"
+	LoadBalancerHTTPMatcherMethodPatch   LoadBalancerHTTPMatcherMethod = "PATCH"
+	LoadBalancerHTTPMatcherMethodDelete  LoadBalancerHTTPMatcherMethod = "DELETE"
+	LoadBalancerHTTPMatcherMethodConnect LoadBalancerHTTPMatcherMethod = "CONNECT"
+	LoadBalancerHTTPMatcherMethodOptions LoadBalancerHTTPMatcherMethod = "OPTIONS"
+	LoadBalancerHTTPMatcherMethodTrace   LoadBalancerHTTPMatcherMethod = "TRACE"
```
ka-myl, line 75: To be frank, I'm not exactly sure if those need to be constants.
HTTP methods are well known and stable, just using plain strings is imo enough and readable. That being said, it doesn't hurt to have them

ka-myl (approved):

upcloud/load_balancer.go
peknur: I thought that this way user can directly see what methods `matcher` supports.

PR #114, opened 2022-03-10
Chore/lb backend requests tests

ka-myl:


upcloud/request/load_balancer.go
```diff
@@ -56,11 +56,11 @@ func (r *DeleteLoadBalancerRequest) RequestURL() string {
 }
 
 type GetLoadBalancerBackendsRequest struct {
-	UUID string `json:"-"`
+	ServiceUUID string `json:"-"`
```
ka-myl, line 59: Even though there is no other `UUID` to pass here, having just `UUID` here was a bit inconsistent with the rest of the request types for load balancer sub-entities (they all use `ServiceUUID`).

peknur: Yep that makes sense. Although I would consider naming it `LoadBalancerUUID ` ? Our API documents do talk about "service" and "services" but in SDK it's still basically `LoadBalancer` type's ID.

peknur (approved):
LGTM

PR #141, opened 2022-04-08
chore(service): refactor service package files

peknur:
Move firewall, server, IP address and tag interfaces and implementations to their own files.


ka-myl (approved):
LGTM

PR #148, opened 2022-04-21
feat(client): new debug flag for setting custom base URL

peknur:
Add new env debug flag `UPCLOUD_DEBUG_API_BASE_URL` to customize API base URL.


ka-myl (approved):
LGTM

PR #157, opened 2022-05-27
fix(firewall): drop IP address family requirement

peknur:
Omit empty IP address family from firewall rule request so that default
rules can be created without family field.


ka-myl (approved):
LGTM

PR #159, opened 2022-06-01
Feat/request context support

peknur:
Adds context-aware client and service types.


ka-myl (approved):
Code looks good. Did some quite intensive testing, no problems found

PR #171, opened 2022-08-17
ci(lint): update lint workflow, enable more linter rules, and gofumpt codebase

kangasta:


villevsv-upcloud (approved):
LGTM, pretty minimal refactoring & cleanup was needed after all!

kangasta:
Yeah, we would likely want to enable `revive` as well at some point, but that would either require disabling `var-naming` rule or tuning the allow and deny lists for variable names, as some of the default naming rules conflict with our current interfaces.

PR #172, opened 2022-09-02
feat(permissions): manage resource permissions

peknur:


kangasta:
Didn't test locally, code looks good to me!

PR #174, opened 2022-09-12
feat(kubernetes): experimental: cluster, plans & versions 

villevsv-upcloud:


upcloud/kubernetes.go
```diff
@@ -0,0 +1,60 @@
...
+type KubernetesMaintenancePeriod struct {
+	DOW  string `json:"dow,omitempty"`
+	Hour *int   `json:"hour,omitempty"`
+}
+type KubernetesPlan struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+}
+
+type KubernetesError struct {
```
peknur, line 51: Is this type needed or can we use `upcloud.Error` ?

upcloud/request/kubernetes.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/UpCloudLtd/upcloud-go-api/v4/upcloud"
+)
+
+const (
+	KubernetesClusterBasePath string = "/kubernetes"
```
peknur, line 11: Does this need to be exported?

upcloud/service/kubernetes.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/UpCloudLtd/upcloud-go-api/v4/upcloud/request"
+)
+
+type Kubernetes interface {
+	CreateKubernetesCluster(r *request.CreateKubernetesClusterRequest) (*upcloud.KubernetesCluster, error)
+	DeleteKubernetesCluster(r *request.DeleteKubernetesClusterRequest) error
+	GetKubernetesCluster(r *request.GetKubernetesClusterRequest) (*upcloud.KubernetesCluster, error)
+	GetKubernetesClusters(r *request.GetKubernetesClustersRequest) ([]upcloud.KubernetesCluster, error)
+	GetKubernetesClustersWithFilters(r *request.GetKubernetesClustersWithFiltersRequest) ([]upcloud.KubernetesCluster, error)
+	GetKubernetesKubeconfig(r *request.GetKubernetesKubeconfigRequest) (*string, error)
```
peknur, line 17: Maybe return just string instead of pointer ?

upcloud/kubernetes.go
villevsv-upcloud: In the API spec there is field `error_details` (string) in the error schema, so created a new type for keeping things a bit cleaner. I can embed the `upcloud.Error`, but let's see if the details is needed in the end.

upcloud/request/kubernetes.go
villevsv-upcloud: Nope! Fixed.

upcloud/service/kubernetes.go
villevsv-upcloud: Fixed.

villevsv-upcloud:
LGTM

PR #180, opened 2022-10-24
test: do not record wait calls

villevsv-upcloud:


upcloud/request/load_balancer.go
```diff
@@ -229,7 +229,7 @@ func (r *DeleteLoadBalancerBackendMemberRequest) RequestURL() string {
 // LoadBalancerResolver represents resolver payload
 type LoadBalancerResolver struct {
 	Name         string   `json:"name,omitempty"`
-	Nameservers  []string `json:"nameservers"`
+	Nameservers  []string `json:"nameservers,omitempty"`
```
peknur, line 232: I think that if we omit empty nameserver slice we are not able to remove nameservers when modifying resolver? For example
```go
t.Logf("Remove resolver %s nameservers", name)
resolver, err = svcContext.ModifyLoadBalancerResolver(ctx, &request.ModifyLoadBalancerResolverRequest{
	ServiceUUID: lb.UUID,
	Name:        resolver.Name,
	Resolver: request.LoadBalancerResolver{
		Nameservers: make([]string, 0),
	},
})
require.NoError(t, err)
assert.Len(t, resolvers, 0)
```

peknur (commented):
Found one possible issue, but otherwise PR looks good!

upcloud/request/load_balancer.go
peknur: Sorry, my bad. I remembered that nameservers could be also empty.

peknur (approved):

upcloud/request/load_balancer.go
villevsv-upcloud: Thanks for being thorough, the documentation indeed states they could be empty but in real life that seems not to be the case. The other option here is to remove `omitempty` (as it was) and let the server return validation error. What do you think is the best?

peknur: That is a good question :)  Normally I prefer to use `omitempty` but slices are sometimes tricky because for example if one would use code above trying to reset nameservers it might be unexpected behavior to not get error. Anyway, as nameserver list can not be empty I would go with `omitempty` . I think that original implementation was created based on docs and that's why that tag was missing.

PR #182, opened 2022-11-03
feat(lbaas): private network support

peknur:
Support for load balancer network objects:
https://developers.upcloud.com/1.3/17-managed-loadbalancer/#networks


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Added
+- Managed Load Balancer private network support
+
+### Deprecated
+- Managed Load Balancer fields `DNSName` and `network_uuid`
```
kangasta, line 12: Could use go field name instead of JSON field name also for `NetworkUUID`

Makefile
```diff
@@ -0,0 +1,19 @@
...
+test: check-test-env fmt vet
+	go test ./... -v -parallel 8
+
+.PHONY: fmt 
+fmt: 
+	go fmt ./...
+
+.PHONY: vet
+vet:
+	go vet ./...
```
kangasta, line 11: We also have `gofumpt` (instead of `gofmt`) and `govet` in the `golangci-lint` configuration, so not sure if these are needed here. One alternative would be to run `golangci-lint` from the `Makefile`.

upcloud/load_balancer_test.go
```diff
@@ -99,6 +99,28 @@ func TestMarshalLoadBalancer(t *testing.T) {
 						}
 					]
 				}
+			],
+			"nodes": [
+            	{
+					"networks": [
+						{
```
kangasta, line 106: Does this omit `uuid` and `dns_name` on purpose?

kangasta (approved):

kangasta:
Looks good to me! Added few non-blocking comments/questions.

CHANGELOG.md
peknur: Yep should be `NetworkUUID`. I'll push update.

Makefile
peknur: I'll remove those targets. I guess idea was to just do some quick tests before running tests but as we have already golangci-lint config and pre-commit setup that's not necessary

upcloud/load_balancer_test.go
peknur: Those nodes are not that well documented yet and I forgot to update those JSON strings in upcloud package after I noticed that there is also other fields :) 
I'll push update.

villevsv-upcloud (approved):
LGTM

peknur:
@villevsv-upcloud thanks for review!

PR #184, opened 2022-11-09
Feat/anti affinity groups

ka-myl:


upcloud/request/server_group.go
```diff
@@ -91,11 +117,23 @@ func (s ModifyServerGroupRequest) RequestURL() string {
...
 	}{}
-	v.ServerGroup = c(r)
+
+	v.ServerGroup = serverGroupPayload{
+		Labels:  r.Labels,
+		Members: r.Members,
+		Title:   r.Title,
+	}
+
+	if !r.AntiAffinity.Empty() {
```
ka-myl, line 130: This logic is (partially) duplicated, because the payload for modify and create requests is actually slightly different - in create `Members` is a slice, in modify `Members` is a pointer to a slice

upcloud/request/server_group.go
```diff
@@ -55,11 +55,21 @@ func (s GetServerGroupRequest) RequestURL() string {
 	return fmt.Sprintf("%s/%s", serverGroupBasePath, s.UUID)
 }
 
+// serverGroupPayload represents the data that we send to server group API.
+// It exists to allow some type transformations for improved user experience.
+type serverGroupPayload struct {
+	Labels       *upcloud.LabelSlice      `json:"labels,omitempty"`
+	Members      *upcloud.ServerUUIDSlice `json:"servers,omitempty"`
+	AntiAffinity string                   `json:"anti_affinity,omitempty"`
```
ka-myl, line 63: The main reason I added this type is because API expects string (`"yes"` or `"no"`) in payload, while I think from the user perspective it's much better to use the `upcloud.Boolean` type

peknur: I think that we don't need that separate payload type because `upcloud.Boolean`'s `MarshalJSON` already does that conversion?

peknur (approved):
LGTM

villevsv-upcloud (approved):
LGTM

PR #186, opened 2022-11-16
Remove old Service type and refactor Client for next v5 release

peknur:
### Changed
- bumped module version from `v4` to `v5`
- context-aware type `client.ClientContext` refactored and renamed to `client.Client`
- context-aware type `service.ServiceContext` renamed to `service.Service`
- `service.Service` accepts any client that implements new `service.Client` interface
- `client.New` constructor allows overwrite `baseURL`, `httpClient` and `httpClient.Timeout` client properties

### Removed
- `service.Service` type without context support
- `client.Client` type without context support
- redundant `Timeout` field from `request.StartServerRequest` type


nate-droid (approved):
Looks good! Nice to see how the context changes removed some boilerplate as well!

upcloud/client/client.go
```diff
@@ -2,216 +2,210 @@ package client
...
 )
 
-// Constants
 const (
-	DefaultAPIVersion = "1.3.6"
-	DefaultAPIBaseURL = "https://api.upcloud.com"
-
-	// The default timeout (in seconds)
-	DefaultTimeout = 60
+	Version    string = "5.0.0"
```
ka-myl, line 20: This replaces the `internal/globals.go` right? If so, we also need to update RELEASING.md to point to this line in step 3.

ka-myl (approved):

upcloud/client/client.go
peknur: @ka-myl Yep, I thought that there is no reason to keep separate package for this.  I'll update RELEASING.md if this is otherwise OK ?

upcloud/service/service.go
```diff
@@ -1,107 +1,98 @@
...
+	Delete(ctx context.Context, path string) ([]byte, error)
+	// Do performs a HTTP request using custom request object and returns the response body.
+	Do(r *http.Request) ([]byte, error)
 }
 
-var _ TimeZones = (*Service)(nil)
-
-type Plans interface {
-	GetPlans() (*upcloud.Plans, error)
+type requestable interface {
```
villevsv-upcloud, line 28: What about calling this `request`?

upcloud/service/utils_test.go
```diff
@@ -53,8 +51,8 @@ func handleError(err error) {
 	}
 }
 
-// records the API interactions of the test.
-func record(t *testing.T, fixture string, f func(*testing.T, *recorder.Recorder, *Service)) {
+// records the API interactions of the test. Function provides both services to test cases so that old utility functions can be used to initialize environment.
+func recordWithContext(t *testing.T, fixture string, f func(context.Context, *testing.T, *recorder.Recorder, *Service)) {
```
villevsv-upcloud, line 55: `WithContext` suffix could probably be removed as obsolete, same for other helper functions with that in their name

villevsv-upcloud (approved):
LGTM! The amount of removed lines makes me happy. Left some non-blocking comments.

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Changed
+- bumped module version from `v4` to `v5`
+- context-aware type `client.ClientContext` renamed to `client.Client`
+- context-aware type `service.ServiceContext` renamed to `service.Serivce`
+- `service.Serivce` accepts any client that implements `service.Client` interface
```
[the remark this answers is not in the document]
peknur: Good catch :) Thanks

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Changed
+- bumped module version from `v4` to `v5`
+- context-aware type `client.ClientContext` renamed to `client.Client`
+- context-aware type `service.ServiceContext` renamed to `service.Serivce`
+- `service.Serivce` accepts any client that implements `service.Client` interface 
+- `client.New` constructor allows overwrite `baseURL`, `httpClient` and `httpClient.Timeout` client properties
+
```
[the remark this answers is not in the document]
peknur: Absolutely, it's now added

upcloud/service/service.go
peknur: I thought also first but I think that then we would need to alias imports from `request` package. We have had this interface for a while and I struggled with the name already then :)  Current name is not perfect but it seems that it is used else where also to name this kind of interface. Anyway it's private and if we come up with better name (other than request) we can change it later.

go.mod
```diff
@@ -1,14 +1,18 @@
-module github.com/UpCloudLtd/upcloud-go-api/v4
+module github.com/UpCloudLtd/upcloud-go-api/v5
 
-go 1.14
+go 1.18
```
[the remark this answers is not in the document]
peknur: Thanks @villevsv-upcloud for review! 

I would stay in 1.18 so that we don't accidentally add any 1.19 features and interfere with lib clients that are still in 1.18.

upcloud/service/utils_test.go
peknur: Good point. Pushed changes,

upcloud/service/load_balancer.go
```diff
@@ -96,284 +94,121 @@ func (s *Service) GetLoadBalancers(r *request.GetLoadBalancersRequest) ([]upclou
...
-	if err != nil {
-		return nil, parseJSONServiceError(err)
-	}
-
-	err = json.Unmarshal(res, &members)
-	return members, err
+	return members, s.get(ctx, r.RequestURL(), &members)
 }
 
 // GetLoadBalancerBackendMember retrieves details of a load balancer balancer backend member.
```
villevsv-upcloud, line 154: Just noticed `balancer balancer` typo here and in couple of other places :)

peknur: Fixed :)

PR #189, opened 2022-12-05
feat(error): add HTTP status code to upcloud.Error type

peknur:


villevsv-upcloud (approved):
Non-blocking proposal: as the `client.Error` [struct contains the field](https://github.com/UpCloudLtd/upcloud-go-api/blob/feat/upcloud-error-status-code/upcloud/client/error.go#L14) `ErrorCode`, maybe it could be renamed to `Status` or `StatusCode` to avoid confusion with `upcloud.Error`'s `ErrorCode`?

PR #193, opened 2022-12-13
feat(lbaas): add `Scheme` property to `http_redirect` action

kangasta:


upcloud/load_balancer.go
```diff
@@ -323,9 +327,10 @@ type LoadBalancerActionHTTPReturn struct {
 	Payload     string `json:"payload,omitempty"`
 }
 
-// LoadBalancerActionHTTPRedirect represents 'http_redirect' action
+// LoadBalancerActionHTTPRedirect represents 'http_redirect' action. Only either Location or Scheme should be defined.
```
villevsv-upcloud, line 330: I was thinking if the SDK itself should enforce only one of "Location" or "Scheme", but perhaps that would be over-engineering. Especially, if other similar cases are currently validated only by the API and not by the SDK.

peknur (approved):

upcloud/request/load_balancer_helpers.go
```diff
@@ -31,6 +31,15 @@ func NewLoadBalancerHTTPRedirectAction(location string) upcloud.LoadBalancerActi
 	}
 }
 
+func NewLoadBalancerHTTPRedirectSchemeAction(scheme upcloud.LoadBalancerActionHTTPRedirectScheme) upcloud.LoadBalancerAction {
```
villevsv-upcloud, line 34: In addition to adding this, maybe the `NewLoadBalancerHTTPRedirectAction` could be renamed to `NewLoadBalancerHTTPRedirectLocationAction` for clarity. Would had to be done in a next major release tho, as it's a breaking change.

villevsv-upcloud (approved):
LGTM, left some non-blocking comments.

upcloud/load_balancer.go
kangasta: Yeah, I think it is better to keep this part simple and rely on errors given by API.

The validation could be done in TF provider and `upctl`. If those end up very similar we could then move the validation here.

PR #208, opened 2023-02-01
Feat/api errors

ka-myl:


upcloud/error_utils.go
```diff
@@ -0,0 +1,87 @@
...
+		return false
+	}
+
+	if uError, ok := err.(*Error); ok {
+		return uError.Status == http.StatusNotFound
+	}
+
+	if uProblem, ok := err.(*Problem); ok {
+		return uProblem.Status == http.StatusNotFound
+	}
```
peknur, line 20: IMHO we should have only one error type which would make this function helper cleaner or even unnecessary.

upcloud/error_utils.go
```diff
@@ -0,0 +1,87 @@
+package upcloud
```
peknur, line 1: I would move these to separate package e.g.  `utils` , `errorutils`

upcloud/error_utils.go
```diff
@@ -0,0 +1,87 @@
...
+	return false
+}
+
+func IsAuthenticationFailedError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if uError, ok := err.(*Error); ok {
+		return uError.Status == http.StatusUnauthorized && uError.ErrorCode == ErrCodeAuthenticationFailed
```
peknur, line 65: Is there other kind of "authentication failed" errors than `http.StatusUnauthorized` ?

upcloud/error_utils.go
```diff
@@ -0,0 +1,87 @@
...
+		return uProblem.Status == http.StatusUnauthorized && problemType == ProblemTypeAuthenticationFailed
+	}
+
+	return false
+}
+
+// GetJsonProblemType gets the meaningful part of json+problem type field
+// json+problem `type` field should be a URL to a page that explains the error
+// for the lack of better alternatives we need to use a fragment of that URL for comparing purposes
+func GetJsonProblemType(err *Problem) string {
```
peknur, line 79: Does this need to be exported?

upcloud/error_utils.go
```diff
@@ -0,0 +1,87 @@
...
+	parts := strings.Split(err.Type, "#")
```
peknur, line 80: We could use `strings.SplitN` here if we want to return everything after first `#`

upcloud/error_codes.go
```diff
@@ -0,0 +1,211 @@
+package upcloud
+
+// Error codes used in upcloud.Error struct
+const (
+	ErrCodeAuthenticationFailed                           = "AUTHENTICATION_FAILED"
```
peknur, line 5: We could type these untyped consts as `string`

upcloud/error_utils.go
ka-myl: I agree, but doing it at this point would probably break backwards compatibility. Not sure how much people used to rely on `Error` and `Problem` types tho.

upcloud/error_utils.go
```diff
@@ -0,0 +1,87 @@
...
+	return false
+}
+
+// List of all error codes that indicate that API object already exists
+var alreadyExistsErrCodes = map[string]struct{}{
+	ErrCodeFirewallRuleExists: {},
+	ErrCodeDBExists:           {},
+	ErrCodeServiceExists:      {},
+	ErrCodeInterfaceExists:    {},
+	ErrCodeTagExists:          {},
```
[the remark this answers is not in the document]
ka-myl: I can't say with absolute certainty that these are all, but those are all I found scanning the API docs and poking the API

upcloud/error_utils.go
ka-myl: Nope :P

upcloud/error_codes.go
ka-myl: Good point

upcloud/error_utils.go
ka-myl: 👍

upcloud/error_utils.go
ka-myl: I guess there aren't, let's just use status code here 👍

upcloud/error_utils.go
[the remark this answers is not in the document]
peknur: I was thinking that what happens to e.g. "server already exists" (just example, dunno what server endpoint returns) here:
`return uError.Status == http.StatusConflict && errCodeOk`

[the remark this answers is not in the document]
ka-myl: It gets ignored :D But to be honest, there isn't a good way of handling that other then checking err codes, as `409` status is also used for other stuff

upcloud/error.go
```diff
@@ -3,35 +3,151 @@ package upcloud
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
+
+	"github.com/UpCloudLtd/upcloud-go-api/v5/upcloud/client"
 )
 
+func NewError(clientErr *client.Error) error {
```
ka-myl, line 11: This kinda has to stay exported because it has to be used in `service` package. We can't define this function in `service` package because it needs to set private fields (like `clientError`).

The alternative is just to move this whole logic back to `parseJSONServiceError` function, it's not that much logic anyway

upcloud/error.go
```diff
@@ -3,35 +3,151 @@ package upcloud
...
+		LegacyError *struct {
+			ErrorCode    string `json:"error_code,omitempty"`
+			ErrorMessage string `json:"error_message,omitempty"`
+		} `json:"error,omitempty"`
+		ProblemType          string                     `json:"type,omitempty"`
+		ProblemTitle         string                     `json:"title,omitempty"`
+		ProblemCorrelationID string                     `json:"correlation_id,omitempty"`
+		ProblemStatus        int                        `json:"status,omitempty"`
+		ProblemInvalidParams []ProblemErrorInvalidParam `json:"invalid_params,omitempty"`
 	}{}
```
ka-myl, line 79: This is a bit hacky. It relies on the `omitempty` json tag to check what we actually got from the API.

[Here is an alternative implementation of this custom unmarshaler](https://github.com/UpCloudLtd/upcloud-go-api/pull/209/files). The logic looks nicer, but the overall solution is convoluted and more limiting IMO.

ka-myl:
Not relevant anymore, we have nicer solution

PR #209, opened 2023-02-06
refactor(errors): custom unmarshaler

ka-myl:
This PR has an alternative approach to unamrshaling unified API error in SDK.
It assumes that by the time we get to unmarshaling, the `Error.clientError` should be already set.
This allows to make the unamrshaling logic a bit nicer and less hacky.
The downside is that we cannot really unmarshal the struct itself - because it needs the `clientError` field to be not nil, and that field is private, thus it cannot be set outside of `upcloud` package. So unmarshaling can only really happen in the `NewError` function, which might not be the end of the world, but it sucks.
Overall I don't like this approach due to ☝️ this limitation, but the logic looks nicer.

### DIFF
--- a/upcloud/error.go
+++ b/upcloud/error.go
@@ -14,14 +14,13 @@ func NewError(clientErr *client.Error) error {
 	}
 
 	ucErr := &Error{}
+	ucErr.Status = clientErr.ErrorCode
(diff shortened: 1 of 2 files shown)


ka-myl:
Not relevant anymore, we have nicer solution

PR #210, opened 2023-02-07
Refactor/errors

ka-myl:
This PR refactors how SDK handles service errors. Long story short - all methods now return `Problem` type


upcloud/problem.go
```diff
@@ -34,8 +41,37 @@ func (p *Problem) Error() string {
...
+//		if problem.MatchesProblemType(upcloud.ProblemTypeResourceAlreadyExists) {
+//			handleStuff(problem)
+//		}
+//
+//		// This might work, but is not guaranteed and not recommended
+//		if problem.Type == upcloud.ProblemTypeResourceAlreadyExists {
+//			handleStuff(problem)
+//		}
+//	}
+func (p *Problem) MatchesProblemType(problemType string) bool {
```
ka-myl, line 62: The main reason why we need this is that the `Problem.Type` can be an URL that might change over time. For example now it's `https://developers.upcloud.com/1.3/errors#ERROR_SOME_ERROR_CODE`, but it might change in the future for operational reasons or something.

Technically we could add that whole URL string to constants and not have this method at all, but then if the actual URL in API response changes the SDK consumer code might stop working without any real warning. So I added this here, this is not really fully future-proof, but it seems a bit more secure to me.

upcloud/problem.go
```diff
@@ -34,8 +41,37 @@ func (p *Problem) Error() string {
...
+	// Type is a URL, we need to extract meaningful fragment from it for comparison purposes
+	if strings.Contains(p.Type, "https://") {
```
peknur, line 64: I think that `strings.HasPrefix()` would be a bit more efficient here.

upcloud/problem.go
```diff
@@ -34,8 +41,37 @@ func (p *Problem) Error() string {
...
+		parts := strings.SplitN(p.Type, "#", 2)
```
peknur, line 65: Maybe use `url.Parse()` and use `Fragment` field here.

upcloud/problem.go
peknur: How about `func (p *Problem) ErrorCode() upcloud.ErrorCode`   ? 
I kind of liked more that naming that you had (if I remember correctly) previously `ErrorCodeResourceAlreadyExists` then we might have that as type `ErrorCode` e.g.:
```go
type ErrorCode string
const (
 ErrorCodeResourceAlreadyExists ErrorCode = "..."
 // ...
)
``` 
and usage would be something like:
```go
p.ErrorCode() == upcloud.ErrorCodeResourceAlreadyExists
```

peknur (commented):
Left some comments but otherwise this is starting to look good to me :+1:

upcloud/problem.go
ka-myl: The reason why I changed that to `ProblemTypeXXX` is because... well problem doesn't have `Code` field or anything. But if we add a method like that then sure.
I guess this might also be a bit more go-proper to compare stuff like this

ka-myl: Done

peknur (approved):

upcloud/problem.go
```diff
@@ -34,8 +42,18 @@ func (p *Problem) Error() string {
...
+		parsedURL, err := url.Parse(p.Type)
+		if err != nil {
+			return ""
+		}
+
+		return strings.Replace(parsedURL.Fragment, "ERROR_", "", 1)
+	}
+
+	// Type is just bare error code string (this happens when error returned from the API is not in the json+problem format)
+	return p.Type
```
villevsv-upcloud, line 58: Would the below suggestion work instead of the HTTPS scheme check?

It's a given that we have to make many assumptions. It could be nice to remove one of those: depending on exact URL scheme.

```suggestion
	// If the type is a URL, we need to extract meaningful fragment from it for comparison purposes
	parsedURL, err := url.Parse(p.Type)
	if err != nil {
		// Type is just bare error code string (this happens when error returned from the API is not in the json+problem format)
		return p.Type
	}

	return strings.Replace(parsedURL.Fragment, "ERROR_", "", 1)
```

villevsv-upcloud (approved):
Left a non-blocking minor suggestion.

upcloud/problem.go
ka-myl: Funny enough, `url.Parse("GROUP_NOT_FOUND")` does not return an error :P Need to also check for empty host or scheme. But overall good suggestion, added!

PR #219, opened 2023-04-04
feat(config): tls config: skip verify for provided client

villevsv-upcloud:
upcloud-go-api consumers may define and provide their own HTTP Client while initializing the client. With this change, the debug environment variable `UPCLOUD_DEBUG_SKIP_CERTIFICATE_VERIFY` is respected regardless of the source of the HTTP Client.

Previously, if a consumer provided a client, the transport was unaffected even if `UPCLOUD_DEBUG_SKIP_CERTIFICATE_VERIFY` was set to `1` and skipping TLS verification was not enabled.

### DIFF
--- a/upcloud/client/client.go
+++ b/upcloud/client/client.go
@@ -148,6 +148,18 @@ func WithBaseURL(baseURL string) configFn {
 	}
 }
 
+func WithInsecureSkipVerify() configFn {
+	return func(c *config) {
+		if c.httpClient != nil {
+			c.httpClient.Transport = &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true, //nolint
+				},
+			}
+		}
+	}
+}
+
 func WithHTTPClient(httpClient *http.Client) configFn {
 	return func(c *config) {
 		c.httpClient = httpClient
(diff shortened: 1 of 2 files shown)


upcloud/client/client.go
```diff
@@ -148,6 +148,18 @@ func WithBaseURL(baseURL string) configFn {
 	}
 }
 
+func WithInsecureSkipVerify() configFn {
+	return func(c *config) {
+		if c.httpClient != nil {
+			c.httpClient.Transport = &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true, //nolint
```
kangasta, line 156: While easy to understand given the function, this could have a comment explaining why `//nolint` is used. Also wondering if `//nolint` now also disables `nolintlint` linter here, which should give error about `//nolint` without comment  😅

villevsv-upcloud: Replaced with `nosec` and addded a comment! Makes more sense now.

upcloud/client/client.go
```diff
@@ -142,32 +142,56 @@ func (c *Client) getBaseURL() string {
...
 		c.baseURL = baseURL
 	}
 }
 
+// WithInsecureSkipVerify modifies the client's httpClient to skip verifying
+// the server's certificate chain and host name. This should be used only for testing.
+func WithInsecureSkipVerify() configFn {
+	return func(c *config) {
+		if c.httpClient != nil { // #nosec G402 // allow setting InsecureSkipVerify to true as explicitly requested
+			c.httpClient.Transport = &http.Transport{
```
kangasta, line 157: Does this now also override other `http.Transport` settings with their default values?

villevsv-upcloud: Ah, good point. Updated to set either `TLSClientConfig` or `InsecureSkipVerify` instead.

kangasta (approved):

PR #227, opened 2023-05-26
chore(server): add constants for NIC models

thevilledev:
Add constants for NIC models as defined in [the API spec](https://developers.upcloud.com/1.3/8-servers/#attributes).


peknur (approved):
Looks good :+1:  Good to have more of these defined.

PR #233, opened 2023-05-31
fix(dbaas): delete obsolete GetManagedDatabaseIndex method

villevsv-upcloud:


peknur:
Couldn't find documentations on this, so I just want to make sure that we don't support patching only one property, e.g.:
```json
{
  "access_control": true
}
```

villevsv-upcloud:
> Couldn't find documentations on this, so I just want to make sure that we don't support patching only one property, e.g.:
> 
> ```json
> {
>   "access_control": true
> }
> ```

Thanks for the comment @peknur. I added a test to update only one of the access control settings at a time, and you were totally correct, so reverted the change.

peknur (approved):

thevilledev (approved):

PR #236, opened 2023-06-02
feat(lbaas) inverse rule matcher

peknur:


upcloud/request/load_balancer_helpers.go
```diff
@@ -205,6 +205,16 @@ func NewLoadBalancerSrcIPMatcher(IP string) upcloud.LoadBalancerMatcher {
 	}
 }
 
+func NewLoadBalancerInverseSrcIPMatcher(IP string) upcloud.LoadBalancerMatcher {
```
kangasta, line 208: Might be bit weird to have this inverse helper only for source IP as all other matchers also accept that parameters, but I guess we can add those if there is demand and it is already possible to create the object with this and modify that field after.

kangasta (approved):

upcloud/request/load_balancer_helpers.go
peknur: @kangasta thanks for the review :+1:  Good point. I created alternative, more generic, helper version in https://github.com/UpCloudLtd/upcloud-go-api/pull/238 .

PR #240, opened 2023-06-07
feat(lbaas): support for controlling health check TLS verify

thevilledev:
- Add support for LB backend member property `health_check_tls_verify`. If enabled, no TLS certificate validation is done for backend members that are using `https` scheme for health checks.
- Fix LB service tests.
  - Use valid CIDR ranges.
  - Change `t.Cleanup()` to `defer` to match how other tests are handling cleanup. Worth noting that `t.Cleanup()` does not work as recorder cancels the test `ctx` as soon as the test has concluded.
  - LB `DELETE` calls were present in fixtures but they were explicitly being passed through by the recorder. Now fixtures and tests are aligned.
- Update LB fixtures.


peknur (approved):
Looks good :+1:

PR #242, opened 2023-06-08
refactor(client): new http.Client init function

peknur:
This replaces  `go-cleanhttp`  dependency with `NewDefaultHTTPClient` and `NewDefaultHTTPTransport` functions. Functions are exported so that client and/or transport can be easily modified by user if needed. 

HTTP client properties should be identical to what was previously, except for `ForceAttemptHTTP2`, which is also set in latest `go-cleanhttp` package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


kangasta (approved):
Not very familiar with these packages, but looks good to me!

peknur:
@kangasta & @villevsv-upcloud thanks for the review :+1:

PR #244, opened 2023-06-15
feat(kubernetes) node group node details 

peknur:
Provide information about node group nodes and add experimental resource
to remove specific node from node group.


upcloud/kubernetes.go
```diff
@@ -17,13 +17,20 @@ const (
 	KubernetesClusterTaintEffectNoExecute        KubernetesClusterTaintEffect = "NoExecute"
 	KubernetesClusterTaintEffectNoSchedule       KubernetesClusterTaintEffect = "NoSchedule"
 	KubernetesClusterTaintEffectPreferNoSchedule KubernetesClusterTaintEffect = "PreferNoSchedule"
+
+	KubernetesNodeStateNodeStateFailed KubernetesNodeState = "failed"
```
kangasta, line 21: `NodeState` twice in the `const` name

peknur: Good catch, thanks :+1: I'll push fixed.

kangasta (approved):

PR #245, opened 2023-06-16
chore: use Go 1.19

thevilledev:
- Use Go 1.19 in `go.mod` and CI actions
- Test against two previous Go major releases (1.19, 1.20) and drop out 1.18
- Remove deprecated linters
- Add make command for `lint` :)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/UpCloudLtd/upcloud-go-api/v6
 
-go 1.18
+go 1.19
 
 require (
 	github.com/davecgh/go-spew v1.1.1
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/UpCloudLtd/upcloud-go-api/v6
 
-go 1.18
```
peknur, line 3: Should we still keep this as minimum supported Go version 1.19?

thevilledev: Thanks, good point! Version 1.21 is not far from being released but at least switch to a supported version for now 👍 I'll update the PR

thevilledev: Done!

peknur (approved):
Looks good, good to get these up to date :+1:

villevsv-upcloud (approved):

PR #247, opened 2023-06-28
feat(kubernetes): node group utility network access

villevsv-upcloud:


upcloud/request/kubernetes.go
```diff
@@ -116,15 +116,16 @@ func (r *GetKubernetesNodeGroupRequest) RequestURL() string {
...
+	Count                int                            `json:"count,omitempty"`
+	Labels               []upcloud.Label                `json:"labels,omitempty"`
+	Name                 string                         `json:"name,omitempty"`
+	Plan                 string                         `json:"plan,omitempty"`
+	SSHKeys              []string                       `json:"ssh_keys,omitempty"`
+	Storage              string                         `json:"storage,omitempty"`
+	KubeletArgs          []upcloud.KubernetesKubeletArg `json:"kubelet_args,omitempty"`
+	Taints               []upcloud.KubernetesTaint      `json:"taints,omitempty"`
+	AntiAffinity         bool                           `json:"anti_affinity,omitempty"`
+	UtilityNetworkAccess bool                           `json:"utility_network_access,omitempty"`
```
kangasta, line 128: Would this default to `false`, if user does not explicitly set this field? If yes, would we want this to default to `true`instead to maintain current behavior as the the default?

villevsv-upcloud: Good point. In this case, API using true as default, we need to for example use a pointer here. So the request struct is now updated to use `*bool` and added some test coverage to see it in action better

kangasta (approved):

PR #249, opened 2023-07-11
fix(server): do not modify boolean values by default

kangasta:


villevsv-upcloud (approved):
Thanks for fixing quickly!

PR #255, opened 2023-08-08
Update LICENSE

thevilledev:


scop (approved):
LGTM, year coincides with the repo transfer carried out in 2017, and it seems there's a stream of commits by UpCloud employees from there on.

PR #257, opened 2023-08-17
fix(kubernetes): modify cluster description

villevsv-upcloud:


kangasta (approved):

upcloud/service/kubernetes_test.go
```diff
@@ -17,12 +17,10 @@ import (
 
 const exampleClusterResponse = `
 {
+	"control_plane_ip_filter":null,
```
kangasta, line 20: I might have left this out on purpose, as API side has `omitempty` defined for this(?), but not should not make a difference.

PR #260, opened 2023-09-21
feat(network): `static_routes` field to router

villevsv-upcloud:


upcloud/network.go
```diff
@@ -292,6 +292,14 @@ type Router struct {
...
 	UUID             string             `json:"uuid"`
 	Labels           []Label            `json:"labels"`
+	StaticRoutes     []StaticRoute      `json:"static_routes"`
+}
+
+// StaticRoute represents a Static route
+type StaticRoute struct {
+	Name    string `json:"name"`
+	Route   string `json:"route"`
+	Nexthop string `json:"nexthop"`
```
peknur, line 302: Should we use NextHop ? :thinking:

peknur (approved):
Looks good, left small open question but it's not a blocker :ok_hand:

ajmyyra (approved):
Looking nice! Only issue I noticed was already commented by Pekka and I do agree that it's not really a blocker.

upcloud/network.go
villevsv-upcloud: I opted to use `Nexthop` for the field (`nexthop`) conversion, as it the original does not contain underscore. I do agree that `NextHop` would "look" better, but I would still keep it as it is to follow previous conventions.

PR #264, opened 2023-09-29
chore: prepare 6.7.1 release

villevsv-upcloud:


kangasta (approved):
I have sometimes just deleted the failed release (and tag) and created it again on the fixing commit, but I believe this is the more correct approach 👌

PR #265, opened 2023-10-10
feat: managed object storage

villevsv-upcloud:


peknur (approved):
Couldn't spot any problems, nice work :+1:

PR #278, opened 2023-11-14
feat(load_balancer): HTTP/2 support & backend TLS config

villevsv-upcloud:


upcloud/load_balancer.go
```diff
@@ -167,8 +168,9 @@ type LoadBalancerFrontendTLSConfig struct {
 
 // LoadBalancerFrontendProperties represents frontend properties
 type LoadBalancerFrontendProperties struct {
-	TimeoutClient        int  `json:"timeout_client,omitempty"`
-	InboundProxyProtocol bool `json:"inbound_proxy_protocol"`
+	TimeoutClient        int   `json:"timeout_client,omitempty"`
+	InboundProxyProtocol *bool `json:"inbound_proxy_protocol,omitempty"`
```
villevsv-upcloud, line 172: This change would introduce a breaking change. I'd see it useful for consistency with `HTTP2Enabled` and backend properties being boolean pointers, but can be reverted if needed.

upcloud/load_balancer.go
```diff
@@ -207,6 +218,10 @@ type LoadBalancerBackendProperties struct {
 	HealthCheckExpectedStatus int                              `json:"health_check_expected_status,omitempty"`
 	StickySessionCookieName   string                           `json:"sticky_session_cookie_name,omitempty"`
 	OutboundProxyProtocol     LoadBalancerProxyProtocolVersion `json:"outbound_proxy_protocol,omitempty"`
+	TLSEnabled                *bool                            `json:"tls_enabled,omitempty"`
+	TLSVerify                 *bool                            `json:"tls_verify,omitempty"`
+	TLSUseSystemCA            *bool                            `json:"tls_use_system_ca,omitempty"`
+	HTTP2Enabled              *bool                            `json:"http2_enabled,omitempty"`
```
peknur, line 224: Do we need pointers when default values are `false` ?

villevsv-upcloud: I suppose we don't. `HealthCheckTLSVerify` could be changed to `bool` as well then, as it also defaults to `false`. But that would be a breaking change. Is changing that worth it, what do you think?

upcloud/load_balancer.go
villevsv-upcloud: This and `HTTP2Enabled` could be just `bool` with `omitempty`, and that would not be a breaking change anymore, unless clients are relying on logic that not setting a value would result in `false` in a modification request for example.

upcloud/load_balancer.go
peknur: As discussed, makes request part problematic so lets keep them as you have proposed :+1:

upcloud/load_balancer.go
peknur: I'm not totally against it, but we could also try to remember do this change next time we have also other breaking changes in queue.

peknur (approved):
Looks good :+1:  Added non blocking comment about `InboundProxyProtocol` .

upcloud/load_balancer.go
villevsv-upcloud: Reverted and added a task for changing this and similar fields for a major release.

peknur (approved):

PR #291, opened 2024-01-25
refactor: remove `Timeout` option from `WaitFor*` methods

kangasta:
This is to remove duplicate timeouts: the `context.Context` parameter passed in to the `WaitFor*` methods might already have an timeout defined. Thus, the `Timeout` field might cause confusing errors when, for example, the timeout defined in `Timeout` field is shorter than timeout defined in the context passed in from e.g. Terraform.

### DIFF
--- a/upcloud/service/retry.go
+++ b/upcloud/service/retry.go
@@ -1,0 +1,49 @@
+package service
+
+import (
+	"context"
+	"time"
+)
+
+type retryConfig struct {
+	interval time.Duration
+	// Inverse the should retry logic. By default, operation is retried until operation returns a value. If inverse is set to true, operation is retried while operation returns a value. This should be used, for example, for waiting until resource is deleted.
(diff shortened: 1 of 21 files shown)


upcloud/service/retry.go
```diff
@@ -0,0 +1,44 @@
...
+	interval: time.Second * 5,
+	inverse:  false,
+}
+
+func retry[T any](ctx context.Context, operation func(int, context.Context) (*T, error), config *retryConfig) (*T, error) {
+	if config == nil {
+		config = &defaultConfig
+	}
+
+	ticker := time.NewTicker(config.interval)
```
kangasta, line 24: Validate that interval is not negative, see [TF acceptance test error](https://github.com/UpCloudLtd/terraform-provider-upcloud/actions/runs/7656289645/job/20864777953#step:5:313)

PR #292, opened 2024-01-25
fix(dbaas): access control fields to pointers

villevsv-upcloud:


upcloud/managed_database.go
```diff
@@ -622,10 +622,10 @@ type ManagedDatabaseUserPGAccessControl struct {
 }
 
 type ManagedDatabaseUserRedisAccessControl struct {
```
kangasta, line 624: Do we need similar changes for other DB types as well? Those seem to not use `omitempty`, though 🤔

villevsv-upcloud: Good point, those are changed  now as well.

kangasta (approved):

PR #298, opened 2024-02-26
feat(objsto2): iam support

villevsv-upcloud:


upcloud/managed_object_storage.go
```diff
@@ -85,11 +82,30 @@ type ManagedObjectStorageNetwork struct {
 
 // ManagedObjectStorageUser represents a user for the Managed Object Storage service
 type ManagedObjectStorageUser struct {
-	AccessKeys       []ManagedObjectStorageUserAccessKey      `json:"access_keys"`
-	CreatedAt        time.Time                                `json:"created_at"`
-	OperationalState ManagedObjectStorageUserOperationalState `json:"operational_state"`
-	UpdatedAt        time.Time                                `json:"updated_at"`
-	Username         string                                   `json:"username"`
+	AccessKeys []ManagedObjectStorageUserAccessKey `json:"access_keys"`
+	Arn        string                              `json:"arn"`
```
kangasta, line 86:
```suggestion
	ARN        string                              `json:"arn"`
```

upcloud/service/managed_object_storage_test.go
```diff
@@ -558,30 +560,285 @@ func TestDeleteManagedObjectStorageUserAccessKey(t *testing.T) {
...
+
+		defer func(uuid string) {
+			err = deleteManagedObjectStorageAndUsers(ctx, svc, uuid)
+			require.NoError(t, err)
+		}(storage.UUID)
+
+		policy, err := svc.CreateManagedObjectStoragePolicy(ctx, &request.CreateManagedObjectStoragePolicyRequest{
+			Name:        "testpolicy",
+			ServiceUUID: storage.UUID,
+			Document:    "%7B%22Version%22%3A%20%222012-10-17%22%2C%20%20%22Statement%22%3A%20%5B%7B%22Action%22%3A%20%5B%22iam%3AGetUser%22%5D%2C%20%22Resource%22%3A%20%22%2A%22%2C%20%22Effect%22%3A%20%22Allow%22%2C%20%22Sid%22%3A%20%22editor%22%7D%5D%7D",
```
kangasta, line 661: Is it possible to pass JSON as input (instead of URL encoded JSON)? That would be more human readable 🤔

kangasta: I guess it has to be URL encoded in API requests and responses, but maybe we could automate the encode/decode part in the Go SDK 🤔

kangasta: Or probably less confusing if do not have automatic conversions 😅 In that case, we could use human readable JSON and encode functions in the tests and examples, but this is something we can also add later

kangasta (approved):
Looks good to me! Added comments about making the policy documents more transparent and found one more non-uppercase ARN abbreviation.

upcloud/service/managed_object_storage_test.go
villevsv-upcloud: I kinda agree on the latest view that it could be a bit confusing in comparison to the API usage and documentation. End user should be able to just use `url.QueryEscape` & `url.QueryUnescape` from `net/url` while populating the request and reading the response. :)

upcloud/managed_object_storage.go
villevsv-upcloud: Fixed

PR #306, opened 2024-03-07
feature(gateway): add VPNaaS support

ka-myl:
This PR adds support for VPNaaS:
- Add gateway plans
- Add `Connections` field and allow to modify it
- Add separate CRUD methods for gateway connections
- Add separate CRUD methods for gateway tunnels


upcloud/gateway.go
```diff
@@ -57,3 +105,82 @@ type GatewayRouter struct {
...
+	Phase2IntegrityAlgorithms []GatewayIPSecIntegrityAlgorithm `json:"phase2_integrity_algorithms,omitempty"`
+	// List of Phase 2 Diffie-Hellman group numbers
+	Phase2DHGroupNumbers []int `json:"phase2_dh_group_numbers,omitempty"`
+}
+
+type GatewayTunnelIPSecAuth struct {
+	Authentication GatewayIPSecAuthType `json:"authentication,omitempty"`
+	// PSK is a user-provided pre-shared key.
+	// Note that this field is only meant to be used when providing API with your pre-shared key; it will always be empty in API responses
+	PSK string `json:"psk,omitempty"`
```
ka-myl, line 176: Alternatively I could remove `PSK` field from here and create a separate type for create/modify request that has both `Authentication` and `PSK` fields.
I went with this option because IMO we already have a lot of types around gateways and the less types we have, the easier it is to use SDK. But I don't have a strong opinion on this.

upcloud/gateway.go
```diff
@@ -30,14 +41,50 @@ const (
...
+	// GatewayFeatureVPN is a Virtual Private Network (VPN) service used to establish an encrypted network connection when using public networks
+	GatewayFeatureVPN GatewayFeature = "vpn"
+
+	GatewayConnectionTypeIPSec GatewayConnectionType = "ipsec"
+
+	GatewayRouteTypeStatic GatewayRouteType = "static"
+
+	GatewayTunnelIPSecAuthTypePSK GatewayIPSecAuthType = "psk"
+
+	GatewayIPSecAlgorithm_aes128gcm16  GatewayIPSecAlgorithm = "aes128gcm16"
```
ka-myl, line 65: Added an underscore in those constants, as I believe it's easier to parse with eyes when the actual algorithm type is separated a bit more from the rest.

kangasta:
Could edit the PR title to e.g. `feat(gateway): add VPNaaS support` so that the commit message of squashed commit will follow conventional commits. This can also be done when merging the PR.

upcloud/gateway.go
kangasta: I think we already use both approaches (different type in `request` package vs. shared type from `upcloud` package) elsewhere as well and this seems quite clean to me 👍

kangasta (approved):

villevsv-upcloud (approved):

PR #317, opened 2024-05-17
chore(service): request parameter msg on json array unmarshal error

villevsv-upcloud:


upcloud/service/service.go
```diff
@@ -64,10 +66,21 @@ func (s *Service) get(ctx context.Context, location string, v interface{}) error
...
 	}
-	return json.Unmarshal(res, v)
+
+	err = json.Unmarshal(res, v)
+	if err == nil {
+		return nil
+	}
+
+	if strings.HasPrefix(err.Error(), "json: cannot unmarshal array") {
+		return errors.Join(err, errors.New("get: request parameters might be incorrect"))
```
kangasta, line 80: Should we provide more hints on how the fix the issue here 🤔 I guess missing UUID would be the most common cause of this error:

```suggestion
		return errors.Join(err, errors.New("get: request parameters might be incorrect, check that required fields, such as UUID, are set to valid values"))
```

kangasta (approved):
LGTM, could provide more details about the possible reason of error 🤔 I added one suggestion.

PR #324, opened 2024-07-09
fix(storage): typo in `StorageEncryptionDataAtRest` constant

peknur:


kangasta (approved):

upcloud/storage.go
```diff
@@ -55,8 +55,8 @@ const (
 	StorageImportStateCancelled  = "cancelled"
 	StorageImportStateCompleted  = "completed"
 
-	StorageEncryptionDataAtReset StorageEncryption = "data-at-rest"
-	StorageEncryptionNone        StorageEncryption = "none"
+	StorageEncryptionDataAtRest StorageEncryption = "data-at-rest"
```
kangasta, line 58: Another option would be to mark the one with typo as deprecated and just add the fixed one in addition to that 🤔

villevsv-upcloud: I would go with the deprecation process, even though it's only a small change.

villevsv-upcloud (approved):

PR #325, opened 2024-07-10
feat(client): add option to log requests and responses

kangasta:


upcloud/client/client.go
```diff
@@ -139,6 +169,50 @@ func (c *Client) getBaseURL() string {
...
+func (c *Client) logRequest(r *http.Request, body []byte) {
+	const authorization string = "Authorization"
+
+	if c.config.logger != nil {
+		headers := r.Header.Clone()
+		if _, ok := headers[authorization]; ok {
+			auth := strings.Split(headers.Get(authorization), " ")
+			// Redact the token part of the Authorization header or the whole value if value has no space found.
+			if len(auth) > 1 {
+				headers.Set(authorization, fmt.Sprintf("%s xxxxx", auth[0]))
```
villevsv-upcloud, line 190: Some form of "redacted" could work here as well.

```suggestion
				headers.Set(authorization, fmt.Sprintf("%s [REDACTED]", auth[0]))
```

upcloud/client/client.go
```diff
@@ -139,6 +169,50 @@ func (c *Client) getBaseURL() string {
...
+			} else {
+				headers.Set(authorization, "xxxxx")
```
villevsv-upcloud, line 192: See above

villevsv-upcloud (approved):

upcloud/client/client.go
kangasta: I chose `xxxxx` to be in sync with [URL.Redacted](https://pkg.go.dev/net/url#URL.Redacted), but `[REDACTED]` would maybe cause less confusion 🤔

kangasta: Let's go with `[REDACTED]` as it provider more information

upcloud/client/client.go
```diff
@@ -139,6 +169,50 @@ func (c *Client) getBaseURL() string {
...
+}
+
+func (c *Client) logRequest(r *http.Request, body []byte) {
+	const authorization string = "Authorization"
+
+	if c.config.logger != nil {
+		headers := r.Header.Clone()
+		if _, ok := headers[authorization]; ok {
+			auth := strings.Split(headers.Get(authorization), " ")
+			// Redact the token part of the Authorization header or the whole value if value has no space found.
```
kangasta, line 188:
```suggestion
			// Redact the token part of the Authorization header or the whole value if there is no space to separate scheme from parameters.
```

PR #349, opened 2024-12-10
feat(partner): add Partner API support

ucmase:
Add support for [Partner API](https://developers.upcloud.com/1.3/23-partner-api/).


villevsv-upcloud (approved):
A change log entry would be good to add, otherwise good to go.

PR #371, opened 2025-06-23
chore(ci): enable dependabot version updates

scop:


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+  # Note: `gomod` intentionally not here, this being a library package
```
scop, line 7: On the other hand if we want this package to "force-pull" compatible updates to dependent apps too instead of leaving it for them to take care of, we could enable `gomod` here as well.

Enabling would have the added benefit that that way we'd be testing more closely with what the lower bounds in go.mod are.

villevsv-upcloud: Good point risen, I would enable `gomod` in this project, too.

villevsv-upcloud (approved):

PR #384, opened 2025-07-16
feat(server): add `Devices` field to server details

kangasta:


villevsv-upcloud (approved):
Looks good to merge once the API returned value is addressed one way or the other.

PR #385, opened 2025-07-24
feat(account): add billing summary

MarinX:
This PR adds billing summary support to the account service.
Follows the API /account/billing/summary/{year-month} endpoint as documented in the account section.

Notes:

-  The `BillingCategory` and `BillingResourceGroup` structs handle the nested JSON structure where categories contain both a total_amount and a nested object (e.g., servers.server, managed_databases.managed_database). If there is a better approach, let me know.
- Some billing categories like `storages` and `networks` are not included as they require specific API response examples. If you can provide with that details, I can implement.

### DIFF
--- a/upcloud/account.go
+++ b/upcloud/account.go
@@ -187,3 +187,69 @@ func (a *AccountList) UnmarshalJSON(b []byte) error {
 	*a = append(*a, v.Accounts.Item...)
 	return nil
 }
+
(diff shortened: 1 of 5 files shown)


MarinX:
Perfect. Thanks for the sample JSON.
Updated with additional billing categories.

kangasta (approved):

PR #401, opened 2025-08-14
chore(deps): bump Go version to v1.24

thevilledev:
As per [Go release policy](https://go.dev/doc/devel/release#policy), Go 1.23 is no longer supported.


villevsv-upcloud:
> Thanks for the bump. `.github/workflows/test.yml` would need update too.

Seems to be done. :) :+1:

PR #411, opened 2025-09-02
feat(network): add dhcp_route_configuration

paketeserrano:
- Added dhcp_routes_configuration attributes inside the ip_network object in the network creation and modification
-  Added tests for mashallng and unmarshaling the objects
- Added integration test

### DIFF
--- a/upcloud/request/network_test.go
+++ b/upcloud/request/network_test.go
@@ -166,20 +166,25 @@ func TestMarshalModifyNetworkRequest(t *testing.T) {
 	}
 
 	expectedJSON := `
-	  {
+	{
 		"network": {
-		  "name": "My private network",
 			"ip_networks": {
-			  "ip_network": [
+			"ip_network": [
 				{
-				  "dhcp": "no",
-				  "dhcp_default_route": "no",
-				  "family" : "IPv4"
+				"dhcp": "no",
+				"dhcp_default_route": "no",
+				"dhcp_routes_configuration": {
+					"effective_routes_auto_population": {
+					"enabled": "no"
+					}
+				},
+				"family": "IPv4"
 				}
-			  ]
-			}
-		  }
-	  }
+			]
+			},
+			"name": "My private network"
+		}
+	}
 	`
 
 	actualJSON, err := json.Marshal(&request)


upcloud/network.go
```diff
@@ -88,15 +88,27 @@ func (s *Interface) UnmarshalJSON(b []byte) error {
...
-	Family           string   `json:"family,omitempty"`
-	Gateway          string   `json:"gateway,omitempty"`
+	Address          string           `json:"address,omitempty"`
+	DHCP             Boolean          `json:"dhcp"`
+	DHCPDefaultRoute Boolean          `json:"dhcp_default_route"`
+	DHCPDns          []string         `json:"dhcp_dns,omitempty"`
+	DHCPRoutes       []string         `json:"dhcp_routes,omitempty"`
+	Family           string           `json:"family,omitempty"`
+	Gateway          string           `json:"gateway,omitempty"`
+	DHCPRoutesConfig DHCPRoutesConfig `json:"dhcp_routes_configuration,omitempty"`
```
kangasta, line 111: Could also use the full _configuration_ word in the naming to be more in sync with the API, but this works as well 🤔 For examples:

```suggestion
	DHCPRoutesConfiguration DHCPRoutesConfiguration `json:"dhcp_routes_configuration,omitempty"`
```

upcloud/network.go
```diff
@@ -88,15 +88,27 @@ func (s *Interface) UnmarshalJSON(b []byte) error {
...
 
+type DHCPRoutesConfig struct {
+	EffectiveRoutesAutoPopulation EffectiveRoutesAutoPopulation `json:"effective_routes_auto_population"`
+}
+
+type EffectiveRoutesAutoPopulation struct {
+	Enabled             Boolean  `json:"enabled"`
+	ExcludeBySource     []string `json:"exclude_by_source,omitempty"`
+	FilterByDestination []string `json:"filter_by_destination,omitempty"`
+	FilterByRouteType   []string `json:"filter_by_route_type,omitempty"`
```
kangasta, line 99: We have in some cases (e.g. `type RouterStaticRouteType string` [here](https://github.com/UpCloudLtd/upcloud-go-api/blob/main/upcloud/network.go#L5)) used custom string types for string where there are known possible values, like we have `router-connected-networks` and/or `static-route` for `exclude_by_source` as well as `user` and/or `service` for `filter_by_route_type`, but have not done that for `family` (that only allows `IPv4`) in networks so I guess either would work here 🤔

kangasta (commented):
Added one more comment about naming, otherwise looks good to me 👍

paketeserrano:
@kangasta I added the new names. Waiting on approval and tests passing

PR #435, opened 2025-10-28
feat(network): add effective_routes and dhcp_effective_routes

paketeserrano:


villevsv-upcloud (changes requested):
A couple of unit tests would be useful.

PR #438, opened 2025-10-31
refactor: continue retrying after transient 5xx errors

kangasta:
Also tunes error message formatting and adds HTTP status to problem messages.

### DIFF
--- a/upcloud/service/retry.go
+++ b/upcloud/service/retry.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/UpCloudLtd/upcloud-go-api/v8/upcloud"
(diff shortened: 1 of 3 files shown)


upcloud/service/retry.go
```diff
@@ -23,22 +28,46 @@ func fillDefaults(c *retryConfig) *retryConfig {
...
+
+	var ucErr *upcloud.Problem
+	if errors.As(err, &ucErr) && ucErr.Status >= 500 {
+		return true
+	}
+
+	return false
+}
+
+func Retry[T any](ctx context.Context, operation func(int, context.Context) (*T, error), config *RetryConfig) (*T, error) {
```
kangasta: Or should we keep this as private function and move all wait-until loops from TF provider and upctl to Go SDK? 🤔

PR #456, opened 2025-12-13
fix: handle firewall field as string or number in billing API

### LINKED ISSUE #449: `firewall` value in API billing summary response seems to be `string`, not `float64`
We use `float64` as type for [firewall](https://github.com/UpCloudLtd/upcloud-go-api/blob/main/upcloud/account.go#L244) cost in billing details. This causes an parsing error because API seems to respond with `string`. Documentation for deprecated version of the endpoint suggests that the type should be float64 (like we have currently defined), but maybe we could use custom unmarshaler (or`interface{}`) for now to avoid errors in https://github.com/UpCloudLtd/upcloud-cli/pull/606 for now 🤔

mgajda:
Fixes #449

## Problem
The billing API returns the `firewall` cost field inconsistently:
- Sometimes as a string: `"0"`
- Sometimes as a number: `0`

This caused JSON unmarshaling errors in upcloud-cli (see UpCloudLtd/upcloud-cli#606).

## Solution
Introduced a new `StringTolerantFloat64` type that handles both string and numeric JSON values during unmarshaling, while always marshaling as a number.

## Changes
- Added `StringTolerantFloat64` type in `upcloud/utils.go`
  - Unmarshals from both JSON strings and numbers
  - Marshals as standard JSON number
  - Provides `.Float64()` method for conversion
- Updated `BillingResourceDetail.Firewall` field to use `StringTolerantFloat64`
- Added comprehensive unit tests:
  - Tests for `StringTolerantFloat64` type itself
  - Tests for billing summary with string firewall value
  - Tests for billing summary with numeric firewall value (backward compatibility)

## Implementation Details
The `StringTolerantFloat64` type follows the same pattern as the existing `Boolean` type, making it reusable for other API fields with similar inconsistencies.

## Testing
All existing tests pass, plus new tests cover:
- ✅ Numeric values (e.g., `123.45`)
- ✅ String values (e.g., `"123.45"`)
- ✅ Zero as number (`0`)
- ✅ Zero as string (`"0"`)
- ✅ Empty string (`""`)
- ✅ Marshaling to JSON
- ✅ Billing summary with string firewall
- ✅ Billing summary with numeric firewall


upcloud/account.go
```diff
@@ -247,3 +248,37 @@ type BillingResourceDetail struct {
 	BillableSizeGiB int     `json:"billable_size_gib,omitempty"`
 	Labels          []Label `json:"labels,omitempty"`
 }
+
+func (brd *BillingResourceDetail) UnmarshalJSON(data []byte) error {
```
kangasta: I modified this to use custom `UnmarshalJSON`: I feel like your original solution with the custom type was cleaner, but I would prefer to keep the types in the exported struct unchanged for now 🤔

kangasta (approved):

PR #501, opened 2026-04-30
feat(server): int64 host id fields, deprecate int fields

villevsv-upcloud:


upcloud/request/page_test.go
```diff
@@ -35,7 +35,7 @@ func TestPage(t *testing.T) {
 	assert.Equal(t, "limit=100&offset=0", p.String())
 	p = &Page{
 		Size:   100,
-		Number: math.MaxInt64, // update to math.MaxInt when available (go 1.17)
+		Number: math.MaxInt,
```
scop: Nit, unrelated?

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Changed
+
+- server, host: host identifier fields to `int64` to avoid JSON unmarshal overflows on 32-bit builds
```
scop: Would this not be a breaking change, and warrant a major version bump?

paketeserrano (approved):

upcloud/request/page_test.go
villevsv-upcloud: :+1: Removed from this PR

CHANGELOG.md
villevsv-upcloud: You are correct, there has been a couple of similar changes for which the major version was not bumped. Instead of doing that here, I updated the PR to add new fields and deprecate the old ones.

paketeserrano (approved):

PR #508, opened 2026-05-20
feat(file-storage): add encrypted field to FileStorage and CreateFileStorageRequest

bramhoven:
The UpCloud File Storage API supports AES-256-GCM encryption at rest via a boolean encrypted field on POST /file-storage. Add the field to the FileStorage response struct and CreateFileStorageRequest; encryption is immutable after creation so Modify/Replace requests are left unchanged.

This is a required feature before implementation can start on https://github.com/UpCloudLtd/terraform-provider-upcloud/issues/994

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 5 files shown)


kangasta (approved):
Thank you for the PR! 🙌 I'll merge this and will create a new release soon.

PR #544, opened 2026-06-23
refactor(client): move logging functions to a new module

kangasta:


paketeserrano (approved):
LGTM. Did you check that the logging works fine with TF?

PR #570, opened 2026-07-30
feat(generator): split output files per tag

villevsv-upcloud:


v9/generator/scripts/generate.sh
```diff
@@ -3,14 +3,41 @@ set -euo pipefail
...
+# Generate common server URL file
+go tool oapi-codegen -config ./generator/config/server-urls.yaml -o ./pkg/upcloud/server_urls.gen.go "$SPEC_FILE"
+
+# Loop through all tags in the spec and generate models and client for each tag
+for tag in $(jq -r '.tags[]?.name' "$SPEC_FILE"); do
+  echo "Generating $tag"
+  TAG_SPEC=$(mktemp)
+  jq --arg tag "$tag" -f ./generator/scripts/filter-spec-by-tag.jq "$SPEC_FILE" > "$TAG_SPEC"
+
+  go tool oapi-codegen -config ./generator/config/models.yaml -include-tags "$tag" -o ./pkg/upcloud/"$tag"_models.gen.go "$TAG_SPEC"
```
paketeserrano, line 29: Note: Will  changing the order of the tags create a big diff? I do not understand the whole process but I have seen the dedup and 'seen' function declaration and I am not sure if, for instance, changing the tag order in the doc would endup with that resource in some other package

v9/pkg/upcloud/upcloud_client.gen.go
```diff
@@ -0,0 +1,12 @@
+// Package upcloud provides primitives to interact with the openapi HTTP API.
+//
+// Code generated by github.com/oapi-codegen/oapi-codegen/v2 version v2.6.0 DO NOT EDIT.
+package upcloud
+
```
paketeserrano, line 5: Is it ok to have this empty? I am mentioniing just in case it is some kind of mistake

villevsv-upcloud: It is ok. Let's remove the redundant tag from the spec.

v9/generator/scripts/generate.sh
villevsv-upcloud: The sorting is handled in a further PR. :+1:

kangasta (approved):

